PR #4, opened 2022-10-19
Refactor FrontendEngine

weili-govtech:
- Ticket: MOL-11508
- Moved className, id, revalidationMode, validationMode to data prop so they can be defined via JSON
- Exposed methods through ref (`getFormState`, `submit`)
- Updated main export
> - Export only FrontendEngine
> - Export typings via separate file
- Update frontend engine stories accordingly


src/components/fields/select/select.tsx
```diff
@@ -1,21 +1,22 @@
...
 	// =============================================================================
 	// CONST, STATE, REFS
 	// =============================================================================
 	const {
 		schema: { id, title, validation, ...otherSchema },
 		name,
 		value,
+		error,
 		onChange,
-		...otherProps
```
weili-govtech, line 18: removed 1otherProps1 because they cannot be used / applicable to this component

src/components/fields/select/select.tsx
```diff
@@ -1,21 +1,22 @@
...
+import { InputSelect } from "@lifesg/react-design-system/input-select";
+import { useEffect, useState } from "react";
 import * as Yup from "yup";
 import { TestHelper } from "../../../utils";
 import { useValidationSchema } from "../../../utils/hooks";
 import { IGenericFieldProps } from "../../frontend-engine";
-import { ISelectRef, ISelectSchema } from "./types";
+import { ISelectSchema } from "./types";
 
-export const Select = React.forwardRef<ISelectRef, IGenericFieldProps<ISelectSchema>>((props, ref) => {
```
weili-govtech, line 9: by default react-hook-form wants to get the ref of each field to interact with it (e.g. through onChange, onBlur, etc), however in our case, not all of our fields are able to apply the ref (e.g. image upload in the future), hence we preemptively avoided this approach and stick to apply the events manually

src/components/fields/textfield/textfield.tsx
```diff
@@ -56,29 +56,24 @@ export const TextField = React.forwardRef<HTMLInputElement, IGenericFieldProps<I
 	// =============================================================================
 	// HELPER FUNCTIONS
 	// =============================================================================
-	const formatInputMode = () => {
-		let { inputMode } = "none" as TInputMode;
-
+	const formatInputMode = (): React.HTMLAttributes<HTMLInputElement>["inputMode"] => {
```
weili-govtech, line 59: note: `inputMode` from JSON schema will take precedence. otherwise we derive the inputMode based on the `type` defined in the JSON schema

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -1,73 +1,72 @@
...
 	const {
-		id,
-		className = "",
-		data,
-		defaultValues,
-		validators,
-		conditions,
-		validationMode,
-		revalidationMode,
+		data: {
```
weili-govtech, line 21: note: move all the props into `data` in order for them to be customisable from JSON schema
however we still have the `className` outside the data prop so devs can use `FrontendEngine` as a styled component.

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -1,73 +1,72 @@
...
+		reValidateMode: revalidationMode,
 		defaultValues: defaultValues,
 		resolver: yupResolver(validationSchema),
 	});
+	const { control, watch, handleSubmit: reactFormHookSubmit, formState } = formMethods;
 
-	// ================================================
+	// =============================================================================
 	// EFFECTS
-	// ================================================
```
weili-govtech, line 44: note: moved `buildFieldsFromSchema` to be called directly in `useEffect` to avoid `useCallback` (unnecessary memoisation)

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -76,38 +75,46 @@ const FrontendEngineInner = (props: IFrontendEngineProps) => {
...
-	const handleOnSubmit = (data: any) => {
-		console.log(data);
-		onSubmit && onSubmit();
-	};
-
-	const formatFormId = () => {
-		let formId = "frontend-engine";
+	// =============================================================================
+	useImperativeHandle<Partial<IFrontendEngineRef>, Partial<IFrontendEngineRef>>(ref, () => ({
```
weili-govtech, line 81: note: these are for OneService use case
- need to check whether form is valid to toggle submit button
- the submit button is outside `FrontendEngine`, there is a need to call submit from outside the form

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -76,38 +75,46 @@ const FrontendEngineInner = (props: IFrontendEngineProps) => {
...
-		<StyledForm
-			id={formatFormId()}
-			data-testid={formatFormId()}
-			className={className}
-			noValidate
-			onSubmit={reactFormHookSubmit(handleOnSubmit)}
-		>
-			{fields}
-		</StyledForm>
+		<FormProvider {...formMethods}>
```
weili-govtech, line 101: note: without refs, `FormProvider` is needed for child components to interact with react-hook-form through `useFormProvider`

PR #8, opened 2022-11-03
Create checkbox component

jeraldlyh:
- Create `checkbox` component and corresponding storybook
- Update logic in both `checkbox` and `multiselect` component to handle empty array which does not trigger `{ required: true }` validation by setting it back to undefined


weili-govtech (changes requested):

src/components/fields/checkbox-group/checkbox-group.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+	// =============================================================================
+	// EFFECTS
+	// =============================================================================
+	useEffect(() => {
+		setFieldValidationConfig(
+			id,
+			Yup.array()
+				.of(Yup.string())
+				.test("is-empty-array", "An option is required", (value) => {
```
weili-govtech, line 33: error message should be derived from required validation config

src/components/fields/checkbox-group/checkbox-group.tsx
```diff
@@ -0,0 +1,98 @@
...
+					if (!value) return true;
+
+					return value.length > 0;
```
weili-govtech, line 36: need to check whether field is required first because this rule should only be applicable if the field is required

src/components/fields/checkbox-group/types.ts
```diff
@@ -0,0 +1,17 @@
...
+	value: string;
+}
+
+export interface ICheckboxGroupSchema
+	extends IFrontendEngineBaseFieldJsonSchema<"checkbox">,
+		Omit<CheckboxProps, TFrontendEngineBaseFieldJsonSchemaKeys> {
+	options: IOption[];
+}
+
+export interface ICheckboxOption extends IOption {
```
weili-govtech, line 15: this is no longer used right?

src/components/fields/multi-select/multi-select.tsx
```diff
@@ -28,7 +28,17 @@ export const MultiSelect = (props: IGenericFieldProps<IMultiSelectSchema>) => {
 	// EFFECTS
 	// =============================================================================
 	useEffect(() => {
-		setFieldValidationConfig(id, Yup.array().of(Yup.string()), validation);
+		setFieldValidationConfig(
+			id,
+			Yup.array()
+				.of(Yup.string())
+				.test("is-empty-array", "An option is required", (value) => {
```
weili-govtech, line 35: same as checkbox-group
- derive error message from required validation config
- this is only applicable if field is required

src/components/fields/checkbox-group/checkbox-group.tsx
```diff
@@ -0,0 +1,98 @@
...
+	// RENDER FUNCTIONS
+	// =============================================================================
+	const renderCheckboxes = () => {
+		return (
+			options.length > 0 &&
+			options.map((option) => (
+				<Label key={option.label}>
+					<StyledCheckbox
+						{...otherSchema}
+						id={id}
```
weili-govtech, line 81: because there are multiple checkboxes, we need to differentiate between each of them, we can add index to the id

weili-govtech (changes requested):

weili-govtech (approved):

PR #10, opened 2022-11-10
Tests

weili-govtech:
- MOL-11619
- Set up unit tests with jest and testing-library
- Created unit tests for frontend engine / yup
- Added unit tests as part of pre-push hook
- Made fixes to frontend engine as they were uncovered while writing unit tests

Update: Includes PR #17, [#18](https://github.com/LifeSG/web-frontend-engine/pull/18), #19


jest/jest.config.js
```diff
@@ -0,0 +1,33 @@
+module.exports = async () => ({
```
weili-govtech, line 1: note: mostly copied over from mol-lib-config

src/__tests__/frontend-engine/yup/custom-conditions.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { YupHelper } from "src/components/frontend-engine/yup";
+import { TestHelper } from "src/utils";
+
+describe("validateUinfin", () => {
```
weili-govtech, line 4: note: test cases are similar to mol-lib-api-contract

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -31,32 +32,40 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
...
-		const subscription = watch((value, { name, type }) => console.log(value, name, type));
-
-		return () => subscription.unsubscribe();
-	}, [watch]);
+	const { watch, handleSubmit: reactFormHookSubmit, getValues } = formMethods;
 
 	// =============================================================================
 	// HELPER FUNCTIONS
 	// =============================================================================
 	useImperativeHandle<Partial<IFrontendEngineRef>, Partial<IFrontendEngineRef>>(ref, () => ({
```
weili-govtech, line 40: note: reworked the handles here to fit oneservice use cases
- the original formState depends on `validationMode` and `revalidationMode` to get updated, e.g. if validationMode is onSubmit, it will only get updated on submission which is not reliable
- hence adding `isValid` method to check for validity of form at any time
- added `getValues` in case there is a need to retrieve entered values at any time

src/utils/test-helper.ts
```diff
@@ -14,4 +14,12 @@ export namespace TestHelper {
 
 		return elementName;
 	};
+
+	export const getError = (fn: () => unknown) => {
```
weili-govtech, line 18: note: helper function to reliably assert errors

src/__tests__/components/frontend-engine/yup/custom-conditions.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { YupHelper } from "src/components/frontend-engine/yup";
+import { TestHelper } from "src/utils";
+
+describe("validateUinfin", () => {
```
weili-govtech, line 4: note: test cases are similar to mol-lib-api-contract because the uinfin check is copied from there

jest/jest.config.js
```diff
@@ -0,0 +1,33 @@
...
+	testEnvironment: "jsdom",
+	testMatch: ["<rootDir>/src/**/__tests__/**/*.spec.[jt]s?(x)"],
+	maxConcurrency: 10,
+	collectCoverageFrom: [
+		"<rootDir>/src/**/*.{js,jsx,ts,tsx}",
+		// Generic exclusions
+		"!**/__tests__/**/*",
+		"!**/stories/**/*",
+		"!**/*{M,m}ock{*,*/**/*}",
+		"!**/*{E,e}num{*,*/**/*}",
```
keithtxw, line 13: Do we need all these?

src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
```diff
@@ -0,0 +1,188 @@
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { useRef } from "react";
+import { FrontendEngine } from "src/components";
+import { IFrontendEngineData, IFrontendEngineProps, IFrontendEngineRef } from "../../../components/types";
```
keithtxw, line 4: when do we decide to use absolute or relative imports?

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -31,32 +32,40 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
...
 	const handleSubmit = (data: TFrontendEngineValues) => {
 		console.log(data); // TODO: remove
 		onSubmit?.(data);
 	};
 
+	// =============================================================================
+	// EFFECTS
+	// =============================================================================
+	useEffect(() => {
+		const subscription = watch((value) => onChange?.(value, checkIsFormValid()));
```
keithtxw, line 64: maybe we check if `onChange` specified instead of using the `?` denotation

src/components/frontend-engine/types.ts
```diff
@@ -19,6 +19,7 @@ import { IYupValidationRule, TRenderRules } from "./yup/types";
 export interface IFrontendEngineProps {
 	className?: string;
 	data?: IFrontendEngineData | undefined;
+	onChange?: (values: TFrontendEngineValues, isValid?: boolean) => unknown | undefined;
```
keithtxw, line 22: `isValid?: boolean | undefined`

src/components/frontend-engine/types.ts
```diff
@@ -49,8 +50,10 @@ export type TRevalidationMode = Exclude<keyof ValidationMode, "onTouched" | "all
 export type TValidationMode = keyof ValidationMode;
 
 export interface IFrontendEngineRef extends HTMLFormElement {
-	/** gets information about the entire form state */
-	getFormState: () => FormState<TFrontendEngineValues>;
+	/** gets form values */
+	getValues: () => TFrontendEngineValues;
+	/** check if form is valid */
+	isValid: () => boolean;
```
keithtxw, line 56: should this be optional as well?

src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
weili-govtech: seems like there is no hard-and-fast rule so I decided to switch everything to relative imports because it appears to be the more standard approach. Along the same line, I updated the vscode `settings.json` to use relative imports by default to standardise it.

and just a wild thought, maybe we can have our own custom linter in the future to catch such issues.

src/components/frontend-engine/frontend-engine.tsx
weili-govtech: just curious, what are the cons of optional chaining a function?

src/components/frontend-engine/types.ts
weili-govtech: not really because they will always be defined in `useImperativeHandle`

jest/jest.config.js
weili-govtech: they are not needed indeed, have removed them

keithtxw (approved):
Abit hard for me to validate as i might not be as clear. I'll tentatively approve for now. We'll revisit if required.

src/components/frontend-engine/frontend-engine.tsx
keithtxw: no obvious cons other than not even running the function if `onChange` is not specified.

keithtxw (approved):

PR #16, opened 2022-12-05
Create chip component

jeraldlyh:
- Create Chip component and corresponding storybook (MOL-11752)
- Add `cursor: not-allowed` style to disabled chip component
- Remove duplicate prop in RadioButton storybook


src/components/fields/chips/chips.tsx
```diff
@@ -0,0 +1,187 @@
...
+			return values;
+		}
+		return Object.fromEntries(values.map((value) => [value, true]));
+	};
+
+	// =============================================================================
+	// EVENT HANDLERS
+	// =============================================================================
+	const handleChange = (text: string): void => {
+		let updatedStateValues = clone(stateValue); // Prevent mutation to original state value
```
weili-govtech, line 93: you will need to use `cloneDeep` because `clone` is doing a shallow clone but there are nested objects and they need to be cloned as well

weili-govtech (changes requested):

weili-govtech (approved):

PR #19, opened 2022-12-14
Update storybook

jeraldlyh:
- Change typing from `number` to `numeric` based on HTML inputMode
- Remove duplicated props
- Standardize id for default components
- Extract common props

### DIFF
--- a/src/stories/3-fields/contact-number/contact-number.stories.tsx
+++ b/src/stories/3-fields/contact-number/contact-number.stories.tsx
@@ -23,7 +23,6 @@ export default {
 	argTypes: {
 		...ExcludeReactFormHookProps,
 		...CommonFieldStoryProps("contact"),
-		"contact-default": { table: { disable: true } },
 		placeholder: {
 			description: "Specifies the placeholder text",
 			table: {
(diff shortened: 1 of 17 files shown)


src/stories/3-fields/contact-number/contact-number.stories.tsx
```diff
@@ -50,6 +38,7 @@ export default {
 			description: "Specifies a default country for the input field",
 			table: {
 				type: {
+					// NOTE: change to explicit??
```
weili-govtech, line 41: actually you can change the controls to `select` and provide `InternationalCallingCodeMap` as the options

src/stories/common.ts
```diff
@@ -14,6 +14,7 @@ export const ExcludeReactFormHookProps = {
 };
 
 export const CommonFieldStoryProps = (fieldType: string) => ({
+	[`${fieldType}-default`]: { table: { disable: true } },
 	fieldType: {
```
weili-govtech, line 18: can help me add props for `validation` and `showIf`? 
ref: https://github.com/LifeSG/web-frontend-engine/commit/7fe491e94182e91f5c40a95a4d34b5a272481118#diff-aa249136a8c16ecec20779866c9ab5c889567972fc97125392130206a90e6a0a

src/stories/3-fields/wrapper/wrapper.stories.tsx
```diff
@@ -28,6 +28,7 @@ export default {
 	},
 	argTypes: {
 		...ExcludeReactFormHookProps,
+		...CommonFieldStoryProps("wrapper"),
```
weili-govtech, line 31: `CommonFieldStoryProps` will add the `label` prop which is not present in `wrapper`, so there is a need to disable that prop in this story

weili-govtech (approved):

PR #20, opened 2022-12-16
Custom validation

weili-govtech:
- MOL-11754
- Updated Yup helper to track and handle custom defined yup conditions
- Added generic to various interfaces so it is possible to extend validation config
- Added generic to frontend engine component as a way to pass the custom config
  - `<FrontendEngine<ICustomValidationRules> />`
- Added `formRef.addCustomValidation()` to add custom validation

### DIFF
--- a/src/components/frontend-engine/frontend-engine.tsx
+++ b/src/components/frontend-engine/frontend-engine.tsx
@@ -93,7 +93,16 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
 	);
 });
 
+/**
+ * prevents inferrence
+ * https://stackoverflow.com/questions/56687668/a-way-to-disable-type-argument-inference-in-generics
+ */
 type NoInfer<T, U> = [T][T extends U ? 0 : never];
+/**
+ * The one and only component needed to create your form
+ *
+ * Minimally you will need to set the `data` props which is the JSON schema to define the form
+ */
 export const FrontendEngine = forwardRef<IFrontendEngineRef, IFrontendEngineProps>((props, ref) => {
 	return (
 		<YupProvider>
@@ -103,3 +112,11 @@ export const FrontendEngine = forwardRef<IFrontendEngineRef, IFrontendEngineProp
 }) as <V = IYupValidationRule>(
 	props: IFrontendEngineProps<IYupValidationRule | NoInfer<V, IYupValidationRule>> & { ref?: Ref<IFrontendEngineRef> }
 ) => ReactElement;
+
+/**
+ * casting as a function with generics as a way to define generics with forwardRef
+ * the generics are a way to set custom definition of Yup validation config
+ * `NoInfer` typing is to ensure validation config is either the default `IYupValidationRule` or explicitly typed
+ * this prevent inferring from the data props entered by devs
+ * can refer to `Add Custom Validation` story for more info
+ */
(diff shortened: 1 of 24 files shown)


src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -90,8 +91,30 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
...
+ * Minimally you will need to set the `data` props which is the JSON schema to define the form
+ */
+export const FrontendEngine = forwardRef<IFrontendEngineRef, IFrontendEngineProps>((props, ref) => {
+	return (
+		<YupProvider>
+			<FrontendEngineInner {...props} ref={ref} />
+		</YupProvider>
+	);
+}) as <V = IYupValidationRule>(
+	props: IFrontendEngineProps<IYupValidationRule | NoInfer<V, IYupValidationRule>> & { ref?: Ref<IFrontendEngineRef> }
```
qroll, line 111: does `V` need to extend `IYupValidationRule`?

src/components/frontend-engine/yup/helper.ts
```diff
@@ -12,8 +12,9 @@ import {
 
 interface IYupCombinedRule extends IYupRenderRule, IYupValidationRule {}
 
-// TODO: custom validation
 export namespace YupHelper {
+	const customYupConditions: string[] = [];
```
qroll, line 16: this is static, so if multiple form engines are rendered this could potentially (1) keep expanding and have duplicates or (2) allow a form engine to access a custom rule it didn't define itself. for (1) a Set would resolve it, for (2) I'm thinking is not a problem because the `IYupValidationRule` typing would restrict this

src/components/frontend-engine/frontend-engine.tsx
weili-govtech: thanks for raising, after consideration, I decided to change the underlying validation config type to be an intersection of `IYupValidationRule` and a couple of other generics (https://github.com/LifeSG/web-frontend-engine/pull/20/files#diff-74758a5fab71a992c8c83c2ef5dec2dcd28b8422aa7002ba6037a3be86289cdaR103)

the consideration is in most cases, the fields are using `IYupValidationRule` and will occasionally need to add a few more rules hence having the intersection will simplify the typings

with this change, all `V` generic defaults to undefined. 

there is a new `U` generic to allow fields to customise validation rules internally, this prevents getting overwritten by the custom validation type.

src/components/frontend-engine/yup/helper.ts
weili-govtech: I cant see a sensible way out of this actually, at its core, it is relying on `Yup.addMethod()` which is also a static function. even if we deconflict the conditions internally, multiple instances of the forms will still create the rules in the same Yup class.

qroll: this may be a problem if user is rendering two forms simultaneously AND adding custom validators with the same name but different logic. but I do feel it's unlikely.

then I think minimally we can dedupe `customYupConditions` using a set

was wondering if it's possible to use `schema.test()` instead, but it seems kind of complicated to use.

weili-govtech: yeah I get what you mean, with `Set` I can dedupe the condition name but there is still an ambiguity in whether it is applied. in view of that, for clarity's sake I added a check against `customYupConditions` to see if the condition is applied already, if yes, it will not be applied again and does a `console.error`

qroll (approved):

PR #23, opened 2022-12-20
Create TextBody and Alert components

jeraldlyh:
- Created `TextBody` component alongside with its corresponding storybook and unit tests (MOL-11908)
- Created `Alert` component alongside with its corresponding storybook and unit tests (MOL-11907)
- Add import for `FrontendEngineElements` to allow `Wrapper` component to render respective elements
- Shift `Wrapper` component from Field to Element
- Create generic interfaces to reduce boilerplate code
- Rename interface to distinguish between `Field` and `Element` schemas

PR #24, opened 2023-01-03
[MOL-11929] Create Text wrapper component

jeraldlyh:
- Delete `TextBody` component
- Create `Text` wrapper component to render text from design system alongside with its corresponding storybook and unit tests
- To review after this [PR](https://github.com/LifeSG/web-frontend-engine/pull/23) is merged


src/stories/4-elements/wrapper/wrapper.stories.tsx
```diff
@@ -34,11 +34,11 @@ export default {
 			description: "Actual HTML element type to render the component as",
 			table: {
 				type: {
-					summary: "div|span|section|header|footer|h1|h2|h3|h4|h5|h6|p",
```
weili-govtech, line 37: understand that the types are removed to prevent conflicting with the new `Text` field, however I think these types are still needed. can we prepend all the field types in `Text` with `text-`?

let's also try to standardise the fieldType naming convention to be kebab-case, e.g. `text-dbody`, `text-h1`, `text-xsmall`

I also noticed the `Various Elements` story is broken, but I think this change will fix the issue

weili-govtech:
can I trouble you rebase onto main first, this way I can see the differences more easily

weili-govtech (changes requested):

weili-govtech (approved):

PR #26, opened 2023-01-11
Publish

weili-govtech:
Misc updates wrt publishing on npm / its use by other projects
- Moved design system to peer dependencies
- Resolved npm audit warning for debug
- Fixed issue with updating form values when default values are updated
- Make types importable in main file
  - i.e. able to import necessary types from `web-frontend-engine` instead of `web-frontend-engine/components/types`
- Updated frontend-engine and contact stories

### DIFF
--- a/src/components/frontend-engine/frontend-engine.tsx
+++ b/src/components/frontend-engine/frontend-engine.tsx
@@ -1,6 +1,7 @@
 import { yupResolver } from "@hookform/resolvers/yup";
 import { forwardRef, ReactElement, Ref, useCallback, useEffect, useImperativeHandle } from "react";
 import { FormProvider, useForm } from "react-hook-form";
+import { useDeepCompareEffectNoCheck } from "use-deep-compare-effect";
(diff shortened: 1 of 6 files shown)


src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -71,6 +72,10 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
 		}
 	}, [checkIsFormValid, onChange, watch]);
 
+	useDeepCompareEffectNoCheck(() => {
+		reset({ ...defaultValues, ...getValues() });
```
weili-govtech, line 76: apparently we have to use `reset` to update the field values when `defaultValues` is changed. unfortunately `reset` will clear all values entered by user hence the solution is to combine both entered values and default values. I am giving the higher precedence for the entered values as they are explicitly defined by the user.

keithtxw (approved):

PR #28, opened 2023-01-12
Publish

weili-govtech:
- this PR also contains commits for soft validation
- replaced @js-joda/locale_en with @js-joda/locale_en-us to reduce package size
- updated wrapper to render elements without controller
  - this prevents them from getting registered in react-hook-form
- added maxlength and derive min / max attributes for numeric field
- added placeholder to chip textarea
- created useValidationConfig hook
  - hook is for formValidationConfig only
  - updated useValidationSchema to handle schema generation only
- updated readme and contributing docs


src/components/elements/wrapper/wrapper.tsx
```diff
@@ -38,31 +38,40 @@ export const Wrapper = (props: IWrapperProps): JSX.Element | null => {
...
-								name={id}
-								shouldUnregister={true}
-								render={({ field, fieldState }) => {
-									const fieldProps = { ...field, id, ref: undefined }; // not passing ref because not all components have fields to be manipulated
-									return <Field schema={child} {...fieldProps} {...fieldState} />;
-								}}
-							/>
-						</ConditionalRenderer>
-					);
+					if (fieldTypeKeys.includes(fieldType)) {
```
weili-govtech, line 49: - differentiate between fields and elements
- only wrap fields in `Controller` to register them into react-hook-form
- elements don't have values so they don't need to be wrapped in `Controller`

src/utils/hooks/use-validation-config.ts
```diff
@@ -0,0 +1,37 @@
+import { useContext, useEffect, useState } from "react";
```
weili-govtech, line 1: created `useValidationConfig` to handle validation config only, the validation schema logic is not part of this hook. this hook is used by all the fields only

the old `useValidationSchema` is updated to store the generated validation schema state only, it is used by the `FrontendEngine` component. this reduces memory usage slightly as the `useEffect` and state updates are not used in the fields anymore.

CONTRIBUTING.md
```diff
@@ -72,30 +38,40 @@ Do also adhere to the guidelines mentioned below.
...
+		│	│	├── index.tsx
+		│	│   └── types.ts
+		│	└── dir-2
+		└── fields
+			├── component-name
+			│   ├── component-name.tsx
+			│   ├── component-name.styles.tsx
+			│	├── index.tsx
+			│   └── types.ts
+			└── dir-2
```
qroll, line 74: updating the markdown formatting

```suggestion
       ├── elements
       │   ├── component-name
       │   │   ├── component-name.tsx
       │   │   ├── component-name.styles.tsx
       │   │   ├── index.tsx
       │   │   └── types.ts
       │   └── dir-2
       └── fields
            ├── component-name
            │   ├── component-name.tsx
            │   ├── component-name.styles.tsx
            │   ├── index.tsx
            │   └── types.ts
            └── dir-2
```

qroll (changes requested):

src/utils/hooks/use-validation-schema.ts
```diff
@@ -1,45 +1,81 @@
...
+					const config = {
+						schema: value.schema,
+						validationRules: value.validationRules ? [...value.validationRules, rule] : [rule],
+					};
+					if (rule.soft) {
+						softValidationConfig = ObjectHelper.upsert<IFieldYupConfig>(softValidationConfig, key, config);
+					} else {
+						hardValidationConfig = ObjectHelper.upsert<IFieldYupConfig>(hardValidationConfig, key, config);
+					}
+				});
```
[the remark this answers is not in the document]
weili-govtech: it appears that `validationRules: value.validationRules ? [...value.validationRules, rule] : [rule]` is causing the duplication, I have decided to remove the forEach loop.

now it will filter the soft / hard validation rules and upsert to the configs depending whether the respective rules exist. this logic will need to be changed further though, I think it can simplified further when we integrate the validation-schema-generator.

qroll (approved):

PR #32, opened 2023-02-14
[MOL-12233][JL] Render API related errors

jeraldlyh:
- Created error handling in frontend engine field
- Created corresponding unit test and storybook


src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -63,18 +73,42 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
...
+		const subscription = watch((value) => {
+			if (onChange) {
 				onChange(value, checkIsFormValid());
-			});
+			}
 
-			return () => subscription.unsubscribe();
-		}
-	}, [checkIsFormValid, onChange, watch]);
+			for (const [key, value] of Object.entries(errors)) {
```
weili-govtech, line 99: this may be a problem in larger forms with huge list of errors, as a quick fix, you can subscribe only if there is `onChange` or the errors contain api errors. be sure to do `getFieldState` only for api errors instead of every error

weili-govtech (changes requested):

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -63,18 +73,42 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
 		onSubmit?.(data);
 	};
 
+	const setErrors = (errors: Record<string, string | string[]>): void => {
+		for (const key in errors) {
+			if (key in fields) {
```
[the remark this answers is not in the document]
jeraldlyh: `getFieldState()` might not be applicable here since the errors has yet to be set on the individual fields

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -63,6 +73,29 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
 		onSubmit?.(data);
 	};
 
+	// NOTE: Wrapper component contains nested fields
+	const setErrors = (errors: TErrorPayload): void => {
+		Object.entries(errors).forEach(([key, value]) => {
+			if (typeof value === "object") {
```
weili-govtech, line 79: need another check to make sure it's not an array

src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
```diff
@@ -188,6 +188,34 @@ describe("frontend-engine", () => {
...
+			try {
+				throw new Error("API error");
+			} catch (error) {
+				ref.current.setErrors({
+					[fieldOneId]: ERROR_MESSAGE,
+				});
+			}
+		};
+
+		it("should support setting of custom errors", async () => {
```
weili-govtech, line 202: as `setErrors()` accepts error message as multiple types like string and array, there is a need to assert the support for each type

src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
```diff
@@ -188,6 +188,34 @@ describe("frontend-engine", () => {
 		expect(getErrorMessage(true)).not.toBeInTheDocument();
 	});
 
+	describe.only("setErrors", () => {
```
weili-govtech, line 191: remove the `.only`

src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
```diff
@@ -188,6 +188,34 @@ describe("frontend-engine", () => {
...
+			}
+		};
+
+		it("should support setting of custom errors", async () => {
+			render(<FrontendEngineWithCustomButton onClick={handleClick} />);
+			await waitFor(() => fireEvent.click(getCustomSubmitButton()));
+
+			expect(getFieldOne().nextSibling.textContent).toMatch(ERROR_MESSAGE);
+		});
+
```
weili-govtech, line 208: can add another test for nested field

weili-govtech (changes requested):

src/components/frontend-engine/frontend-engine.tsx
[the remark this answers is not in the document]
weili-govtech: I am actually referring to the check in `key in fields` but I see that you have created an object helper util function, so it's not an issue already

src/utils/hooks/use-previous.tsx
```diff
@@ -0,0 +1,11 @@
+import { useEffect, useRef } from "react";
```
weili-govtech, line 1: this should be a `.ts` file instead of `.tsx` because it does not contain any HTML elements

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -74,8 +112,25 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
...
+			Object.keys(apiErrors).forEach((key) => {
+				const oldValue = oldFormValues[key];
+				const updatedValue = formValues[key];
+
+				if (oldValue !== updatedValue) {
+					clearErrors(key);
+				}
+			});
+		}
+	}, [formValues, oldFormValues]);
```
weili-govtech, line 132: can remove `oldFormValues` as a dependency for 2 reasons:
- `oldFormValues` is updated when `formValues` change so there is no point checking `oldFormValues`
- `oldFormValues` is a `ref` which does not trigger useEffect

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -38,7 +39,20 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
...
+		watch,
+		handleSubmit: reactFormHookSubmit,
+		getValues,
+		setError,
+		control,
+		formState: { errors },
+		clearErrors,
+	} = formMethods;
+
+	const formValues = useWatch({ control });
```
weili-govtech, line 54: there is a need to disable `useWatch` if there is nothing to watch so it is not watching unnecessarily

actually the previous implementation to watch only if there are api errors is fine too

weili-govtech (changes requested):

src/components/frontend-engine/frontend-engine.tsx
jeraldlyh: noticed that `watch` only unsubscribes upon unmount, hence decided to stick with `useWatch` to disable it dynamically

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -74,8 +112,26 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
...
+			Object.keys(apiErrors).forEach((key) => {
+				const oldValue = oldFormValues[key];
+				const updatedValue = formValues[key];
+
+				if (oldValue !== updatedValue) {
+					clearErrors(key);
+				}
+			});
+		}
+	}, [formValues, errors, watch]);
```
weili-govtech, line 133: - can remove `watch`
- fix `errors` to ensure it triggers useEffect update

weili-govtech (changes requested):

src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -46,13 +46,11 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
...
 		setError,
-		control,
-		formState: { errors },
+		formState,
 		clearErrors,
 	} = formMethods;
 
-	const formValues = useWatch({ control, disabled: !errors });
-	const oldFormValues = usePrevious(formValues);
+	const [oldFormValues, setOldFormValues] = useState<TFrontendEngineValues>({});
```
weili-govtech, line 53: ideally a `useRef` will be better because it will never trigger a rerender. however it is still fine now because the state update happens on a formState update so it is already rerendering.

weili-govtech (approved):

PR #33, opened 2023-02-22
[SUBMIT][WL] add "invalid-form" to submit button's disabled attribute

weili-govtech:
- MOL-12264
- Added "invalid-form" to disable submit button if form is invalid
- Ready to merge
- Keep branch

### DIFF
--- a/src/components/fields/submit-button/submit-button.tsx
+++ b/src/components/fields/submit-button/submit-button.tsx
@@ -1,7 +1,9 @@
 import { Button } from "@lifesg/react-design-system/button";
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
+import { useWatch } from "react-hook-form";
+import useDeepCompareEffect from "use-deep-compare-effect";
 import * as Yup from "yup";
-import { useValidationConfig } from "../../../utils/hooks";
+import { useValidationConfig, useValidationSchema } from "../../../utils/hooks";
 import { IGenericFieldProps } from "../../frontend-engine";
 import { ISubmitButtonSchema } from "./types";
 
@@ -10,11 +12,14 @@ export const SubmitButton = (props: IGenericFieldProps<ISubmitButtonSchema>) => 
 	// CONST, STATE, REF
 	// =============================================================================
 	const {
-		schema: { label, ...otherSchema },
+		schema: { label, disabled, ...otherSchema },
 		id,
 		...otherProps
 	} = props;
 	const { setFieldValidationConfig } = useValidationConfig();
+	const { hardValidationSchema } = useValidationSchema();
+	const formValues = useWatch({ disabled: disabled !== "invalid-form" });
+	const [isDisabled, setIsDisabled] = useState(disabled === true);
 
 	// =============================================================================
 	// EFFECTS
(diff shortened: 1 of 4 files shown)


src/components/fields/submit-button/types.ts
```diff
@@ -3,4 +3,6 @@ import { IFrontendEngineBaseFieldJsonSchema, TComponentOmitProps } from "../../f
 
 export interface ISubmitButtonSchema
 	extends Omit<IFrontendEngineBaseFieldJsonSchema<"submit">, "validation">,
-		TComponentOmitProps<ButtonProps> {}
+		TComponentOmitProps<ButtonProps, "disabled"> {
+	disabled?: boolean | "invalid-form" | undefined;
```
weili-govtech, line 7: decided to add to the `disabled` flag instead of creating a new attribute because the latter option will create ambiguity in behaviour, e.g. should a button stay disabled if `disabled=false` and new attribute defines the button to be disabled if form is invalid?

qroll: I'm personally ok with this, have seen similar options in the wild, where false means always enabled (or disabled) and true means always the opposite, with additional typings for custom behaviour

src/components/fields/submit-button/submit-button.tsx
```diff
@@ -24,11 +29,22 @@ export const SubmitButton = (props: IGenericFieldProps<ISubmitButtonSchema>) =>
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, []);
 
+	useDeepCompareEffect(() => {
+		if (disabled === "invalid-form") {
+			try {
+				hardValidationSchema.validateSync(formValues);
```
qroll, line 35: I might be mixing it up with formik but isn't there a way to derive the current form state from the form context?

weili-govtech: yeah I found that it is possible to check via `isValid` in `useFormContext` but it is a getter and appears to be making updates under-the-hood. I found out because it is throwing a bunch of ` act(...)` warnings in the test cases.

so in view of that, I think manually validating the schema is better and faster.

src/components/fields/submit-button/submit-button.tsx
```diff
@@ -24,11 +29,22 @@ export const SubmitButton = (props: IGenericFieldProps<ISubmitButtonSchema>) =>
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, []);
 
+	useDeepCompareEffect(() => {
```
qroll, line 32: not very familiar with `useWatch`, why is a deep compare needed? (I did notice the object reference of `formValues` would change even when the same dropdown value was selected, but inputs were fine)

qroll (changes requested):
can also trouble you to include a storybook example for this feature?

src/components/fields/submit-button/submit-button.tsx
weili-govtech: hmm I am doing a deep compare to avoid re-rendering the button when the values are touched but not updated, like the dropdown interaction you cited

weili-govtech:
oh yes have just added the stories and unit tests

qroll (approved):

PR #34, opened 2023-02-24
Chips textarea

weili-govtech:
- MOL-12331
- updated chips textarea id to be unique
- format: `<chips id>-textarea-<textarea-label>`
- misc rename to lowercased `textarea`

### DIFF
--- a/src/components/fields/chips/chips.tsx
+++ b/src/components/fields/chips/chips.tsx
@@ -72,7 +72,7 @@ export const Chips = (props: IGenericFieldProps<IChipsSchema>) => {
 		return stateValue.includes(text);
 	};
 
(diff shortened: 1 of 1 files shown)


qroll (commented):
is the lowercasing to `textarea` for schema-related fields only? the component itself is still `TextArea`

weili-govtech:
for `fieldType`, it should be kebab-cased but for id, we have not come up with a standard yet so I continued with kebab-case, are you thinking to adhere to BEM?

weili-govtech:
actually I wanted to adhere to the casing in the design-system, so it should be `Textarea` in the codes, ok let me fix the casing for that component in another commit

qroll (approved):

PR #38, opened 2023-03-09
Update field values on updating schema

weili-govtech:
- MOL-12427
- update fields to check against options and remove values when the corresponding option is removed
	- checkbox-group
	- chips
	- multi-select
	- radio-button
	- select
- add `setValue` method to FrontendEngine
- test cases will be added after the PR is approved

### DIFF
--- a/src/components/fields/checkbox-group/checkbox-group.tsx
+++ b/src/components/fields/checkbox-group/checkbox-group.tsx
@@ -52,12 +52,7 @@ export const CheckboxGroup = (props: IGenericFieldProps<ICheckboxGroupSchema>) =
 	}, [validation]);
 
 	useDeepCompareEffect(() => {
-		const updatedValues = [];
-		options.forEach((option) => {
-			if (value?.includes(option.value)) {
-				updatedValues.push(option.value);
-			}
-		});
+		const updatedValues = value?.filter((v) => options.find((option) => option.value === v));
 		setValue(id, updatedValues);
 	}, [options]);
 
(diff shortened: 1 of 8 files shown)


src/components/fields/checkbox-group/checkbox-group.tsx
```diff
@@ -48,6 +51,16 @@ export const CheckboxGroup = (props: IGenericFieldProps<ICheckboxGroupSchema>) =
...
 	}, [validation]);
 
+	useDeepCompareEffect(() => {
+		const updatedValues = [];
+		options.forEach((option) => {
+			if (value?.includes(option.value)) {
+				updatedValues.push(option.value);
+			}
+		});
+		setValue(id, updatedValues);
```
weili-govtech, line 61: using `setValue` rather than `onChange` to prevent triggering validation
for more complex use cases, devs should use the `setValue()` method exposed through `FrontendEngine` instead

src/components/fields/chips/chips.tsx
```diff
@@ -61,6 +62,16 @@ export const Chips = (props: IGenericFieldProps<IChipsSchema>) => {
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [validation]);
 
+	useDeepCompareEffect(() => {
+		const updatedValues = [];
+		options.forEach((option) => {
+			if (value?.includes(option.value)) {
+				updatedValues.push(option.value);
+			}
+		});
```
qroll, line 71: I feel that it would be more intuitive to act on the values e.g. `value.filter(() => options.find())`

qroll:
I recall from code review that story examples will be omitted, but what about tests?

src/components/fields/chips/chips.tsx
weili-govtech: ok have updated

weili-govtech:
yeah there will be test cases for each of the fields / frontend engine, but I am looking to make a PR after this one to keep it small, do you prefer to review the test cases in this PR too?

qroll:
oh in that case a different PR is fine!

qroll (approved):

PR #49, opened 2023-04-20
Disabled field for radio options

jhyeep:
**Changes**
Added optional 'disabled' field for individual options in Radio component

-   [delete] branch

### DIFF
--- a/src/components/fields/radio-button/types.ts
+++ b/src/components/fields/radio-button/types.ts
@@ -4,7 +4,7 @@ import { IFrontendEngineBaseFieldJsonSchema, TComponentOmitProps } from "../../f
 interface IOption {
 	label: string;
 	value: string;
-	disabled?: boolean;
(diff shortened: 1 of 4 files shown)


src/components/fields/radio-button/types.ts
```diff
@@ -4,6 +4,7 @@ import { IFrontendEngineBaseFieldJsonSchema, TComponentOmitProps } from "../../f
 interface IOption {
 	label: string;
 	value: string;
+	disabled?: boolean;
```
weili-govtech, line 7: can type it as `boolean | undefined`

weili-govtech (changes requested):

weili-govtech:
just to keep the fields consistent, can you update `checkbox` and `chips` with the same `disabled` option? they can be in a separate PR.

weili-govtech (approved):

PR #53, opened 2023-04-28
[VAL][WL] change default validationMode to onTouched

weili-govtech:
**Changes**
- Changed default `validationMode` from `onSubmit` to `onTouched`
- Delete branch

**Changelog entry**
-   Changed default `validationMode` to `onTouched`

**Additional information**
- Reference of `onTouched` behaviour: https://react-hook-form.com/api/useform/#props
- Context: The `onTouched` behaviour is the behaviour that all our forms are using, have checked with UX and various teams regarding this

### DIFF
--- a/src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
+++ b/src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
@@ -384,6 +384,21 @@ describe("frontend-engine", () => {
 			await waitFor(() => fireEvent.change(getFieldOne(), { target: { value: "h" } }));
 			expect(getErrorMessage()).toBeInTheDocument();
 		});
+
+		it("should support onTouched validationMode", async () => {
+			renderComponent(undefined, { validationMode: "onTouched" });
+
+			expect(getErrorMessage(true)).not.toBeInTheDocument();
+
+			await waitFor(() => fireEvent.change(getFieldOne(), { target: { value: "h" } }));
+			expect(getErrorMessage(true)).not.toBeInTheDocument();
+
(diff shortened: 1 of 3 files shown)


src/stories/2-frontend-engine/frontend-engine.stories.tsx
```diff
@@ -220,8 +221,25 @@ ValidateOnBlur.args = {
 	},
 };
 
-export const OnChange: Story<IFrontendEngineProps> = (args: IFrontendEngineProps) => <FrontendEngine {...args} />;
+export const ValidateOnTouched = Template.bind({});
```
qroll, line 224: this can be the default story now, and the old default will be a `ValidateOnSubmit` story

src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
qroll, line 1: the `onSubmit` test case should also fire the blur event on field one to avoid false negatives

src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
```diff
@@ -384,6 +384,31 @@ describe("frontend-engine", () => {
 			await waitFor(() => fireEvent.change(getFieldOne(), { target: { value: "h" } }));
 			expect(getErrorMessage()).toBeInTheDocument();
 		});
+
+		it("should support onTouched validationMode", async () => {
```
qroll, line 388: this is now the default scenario

qroll (changes requested):

qroll (approved):

PR #63, opened 2023-05-17
Event listeners

weili-govtech:
**Changes**
- Implement event listeners to frontend engine
- Delete branch

**Additional information**
- Implementation is from `image-upload` branch
- This is to be used by `image-upload` and `location-input` to fire LifeSG-specific events
- No stories added; they will be added for fields using events instead

### DIFF
--- a/src/components/frontend-engine/event/context-provider.tsx
+++ b/src/components/frontend-engine/event/context-provider.tsx
@@ -1,0 +1,23 @@
+import { createContext, ReactElement, useRef } from "react";
+
+/**
+ * events are added / removed / dispatched via the eventManager
+ * the eventManager is just a div, the handling of events is purely through native event handlers
+ */
+interface IEventContext {
+	eventManager: Element;
+}
+
+interface IProps {
+	children: ReactElement;
+}
+
+export const EventContext = createContext<IEventContext>({
+	eventManager: null,
+});
+
+export const EventProvider = ({ children }: IProps) => {
+	const eventManager = useRef<Element>(document?.createElement("div"));
+
+	return <EventContext.Provider value={{ eventManager: eventManager.current }}>{children}</EventContext.Provider>;
+};
--- a/src/components/frontend-engine/event/context-provider.tsx
+++ b/src/components/frontend-engine/event/context-provider.tsx
@@ -1,4 +1,4 @@
-import { createContext, ReactElement, useRef } from "react";
+import { ReactElement, createContext, useRef } from "react";
 
 /**
  * events are added / removed / dispatched via the eventManager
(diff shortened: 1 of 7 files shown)


src/components/frontend-engine/event/context-provider.tsx
```diff
@@ -0,0 +1,23 @@
...
+interface IProps {
+	children: ReactElement;
+}
+
+export const EventContext = createContext<IEventContext>({
+	eventManager: null,
+});
+
+export const EventProvider = ({ children }: IProps) => {
+	const eventManager = useRef<Element>(document?.createElement("div"));
```
qroll, line 20: instead of creating an element, there seems to be an event target api https://developer.mozilla.org/en-US/docs/Web/API/EventTarget

src/utils/hooks/use-field-event.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Hook that interacts with the event context provider
+ * use this hook to add/dispatch/remove event listeners
+ */
+export const useFieldEvent = () => {
+	const { eventManager } = useContext(EventContext);
+
+	const addFieldEventListener = (
+		type: string,
+		id: string,
+		listener: (this: Element, ev: Event) => unknown,
```
qroll, line 14: do functions get auto bound to `this`?

src/components/frontend-engine/event/context-provider.tsx
qroll: ah oops. it's not available in older browsers like Safari 13 https://caniuse.com/mdn-api_eventtarget_eventtarget

src/utils/hooks/use-field-event.ts
weili-govtech: it doesnt appear to change where `this` is bounded to and is just meant to specify what sort of element the event listener is meant to be added to (`HTMLDivElement` in this case) 

e.g.
```
const addFieldEventListener = (
	type: string,
	id: string,
	listener: (this: string, ev: Event) => unknown,
	options?: boolean | AddEventListenerOptions
) => {
	// some code
};

myElement.addFieldEventListener("myEvent", "fieldId", function (e) {
  // `this` is a string because it is typed as such
});
```

with this, I realise we don't need to type it:
- we usually use arrow functions so `this` is meaningless
- we don't need to refer to the eventManager and are meant to interact with it through the hooks

in view of that, I have removed the typing

src/utils/hooks/use-field-event.ts
```diff
@@ -0,0 +1,36 @@
...
+		options?: boolean | AddEventListenerOptions
+	) => {
+		eventManager?.addEventListener(`${id}:${type}`, listener, options);
+	};
+
+	const removeFieldEventListener = (
+		type: string,
+		id: string,
+		listener: (ev: Event) => unknown,
+		options?: boolean | AddEventListenerOptions
```
qroll, line 24:
```suggestion
		options?: boolean | EventListenerOptions
```

qroll (changes requested):

qroll (approved):

PR #67, opened 2023-05-23
new ignoreDefaultValues prop for reset button

jhyeep:
**Changes**
Added an optional prop `ignoreDefaultValues` for reset-button component that allows users to clear form fields while bypassing any defined default values

-   [delete] branch

**Changelog entry**

-   Added optional ignoreDefaultValues prop for reset button

### DIFF
--- a/src/components/fields/reset-button/reset-button.tsx
+++ b/src/components/fields/reset-button/reset-button.tsx
@@ -8,19 +8,28 @@ export const ResetButton = (props: IGenericFieldProps<IResetButtonSchema>) => {
 	// CONST, STATE, REF
 	// =============================================================================
 	const {
-		schema: { label, disabled, ...otherSchema },
+		schema: { label, disabled, ignoreDefaultValues, ...otherSchema },
 		id,
 		...otherProps
 	} = props;
 
-	const { reset } = useFormContext();
+	const { reset, getValues } = useFormContext();
 
 	// =============================================================================
 	// EFFECTS
 	// =============================================================================
-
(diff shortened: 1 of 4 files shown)


src/components/fields/reset-button/reset-button.tsx
```diff
@@ -8,19 +8,28 @@ export const ResetButton = (props: IGenericFieldProps<IResetButtonSchema>) => {
...
 	const onClick = () => {
-		reset();
+		if (ignoreDefaultValues) {
+			const values = getValues();
+			Object.entries(values).forEach(([key, value]) => {
+				if (Array.isArray(value)) {
+					values[key] = [];
+				} else {
+					values[key] = "";
+				}
```
weili-govtech, line 29: set to empty string if it is of string or number type, otherwise don't update the value

src/stories/3-fields/reset-button/reset-button.stories.tsx
```diff
@@ -51,6 +51,19 @@ export default {
...
+			description: "Specifies if button should reset all fields and default values to blank",
+			table: {
+				type: {
+					summary: "boolean",
+				},
+				defaultValue: { summary: false },
+			},
+			options: [true, false],
+			control: {
+				type: "select",
```
weili-govtech, line 64:
```suggestion
				type: "boolean",
```

weili-govtech (changes requested):

weili-govtech (approved):

PR #72, opened 2023-05-26
Error state

weili-govtech:
**Changes**
- Updated to use `input-textarea` to ensure correct aria-invalid attribute is used
- Update components to show error state


qroll (approved):
noticed that time field does not have a validation story. not sure if it's intended since it only has `required` validation?

otherwise the changes in this PR look ok to me

qroll:
not expecting any action for this in the near future but we should probably re-evaluate aria markup on DS and FEE. in particular [wai-aria 1.2](https://www.w3.org/TR/wai-aria-1.2/#aria-invalid) indicates that aria-invalid is only allowed on specific roles

PR #74, opened 2023-05-29
Allow FEE to take in onSubmitError handler prop

rgchua:
**Changes**
- Expose an onSubmitError prop in FrontendEngine, to be invoked when form submission fails with validation errors
- Wrapped textarea with a div in order to patch an expected behaviour (setting aria-invalid=true attribute when field contains error(s)

-  delete branch

**Changelog entry**
- Allow FrontendEngine to take an onSubmitError prop, to be invoked when form submission fails with validation errors

**Additional information**
-   You may refer to this [RBS-440](https://jira.ship.gov.sg/browse/RBS-440)


src/components/fields/textarea/textarea.tsx
```diff
@@ -89,19 +89,22 @@ export const Textarea = (props: IGenericFieldProps<ITextareaSchema>) => {
...
-					data-testid={TestHelper.generateId(id, "textarea")}
-					name={name}
-					maxLength={maxLength}
-					rows={rows}
-					resizable={resizable}
-					onChange={handleChange}
-					value={stateValue}
-					errorMessage={error?.message}
-				/>
+				<div aria-invalid={!!error?.message}>
```
qroll, line 92: you can take in the latest changes from master. aria-invalid is now applied on the element

src/components/frontend-engine/types.ts
```diff
@@ -45,6 +45,7 @@ export interface IFrontendEngineProps<V = undefined> {
 	onChange?: ((values: TFrontendEngineValues, isValid?: boolean | undefined) => unknown) | undefined;
 	/** Submit event handler, will receive the form data if form validation is successful */
 	onSubmit?: (values: TFrontendEngineValues) => unknown | undefined;
+	onSubmitError?: (errors: TFrontendEngineValues) => unknown | undefined;
```
qroll, line 48: to add a jsdoc comment to describe this property

src/stories/2-frontend-engine/on-submit-error/on-submit-error.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+	const ref = useRef<IFrontendEngineRef>();
+
+	return (
+		<>
+			<FrontendEngine
+				data={{ ...DATA }}
+				ref={ref}
+				onSubmitError={(e) => {
+					console.log("Validation Errors: ", e);
+					const invalidElement = document.querySelector("div:has(*[aria-invalid=true]) > label");
```
qroll, line 125: note that support for `:has` still seems to be limited to newer browsers  https://caniuse.com/css-has

qroll (changes requested):

src/components/fields/textarea/textarea.tsx
rgchua: rebased onto master, no longer manually adding div with aria-invalid attribute

src/components/frontend-engine/types.ts
rgchua: done

src/stories/2-frontend-engine/on-submit-error/on-submit-error.stories.tsx
rgchua: any suggestions? the intention in the sample was to scroll to the label of the associated input containing validation errors. is there any possibility the aria attribute can be applied to the input (textarea) directly such that we can query directly for the label with the right `for` attribute?

`
	const invalidElement = document.querySelector("*[aria-invalid=true]");
					if(invalidElement) {
						const scrollToElement = document.querySelector(`*[for=${invalidElement.id}]`) || invalidElement;
						scrollToElement.scrollIntoView();
					}
`

qroll: not ideal, but
1. there's a `data-testid` set on error messages. if no one overrides `data-error-testid` from the schema, then the `data-testid` value is `<field id>-error-message`. get the element with `[data-testid$="error-message]"` and then scroll to its parent node or label sibling
2. just let it not work if it's not a critical feature and the browser user base is small?

weili-govtech: imo the scrolling behaviour will be up to how an integration service handles so it depends on its use case will be, anyway here are some observations
- there is a need to update the fallback to the selector because `*[for=${invalidElement.id}]` will throw an error if invalidElement has no id (happens to radio buttons)
```
console.log("Validation Errors: ", e);
const invalidElement = document.querySelector("*[aria-invalid=true]");
if (invalidElement && invalidElement.id) {
	document.querySelector(`label[for=${invalidElement.id}]`).scrollIntoView();
} else {
	document.querySelector(`*[aria-invalid=true]`).scrollIntoView();
}
```
- `data-testid` can work unless it defined in the schema indeed, it can work depending on its use case
- unavailability of `:has()` pseudo-element can be overcome with polyfills (disclaimer: I have not tested it)
	- https://www.npmjs.com/package/css-has-pseudo
	- https://github.com/jplhomer/polyfill-css-has

all in all, I think it's not an issue with FEE but rather a concern with integration, i.e. ymmv depending on how you use the library. as a safeguard, we can leave a comment in this story to say the `:has()` pseudo-element is not supported in Firefox yet so use with caution.

weili-govtech: can we also move this story to the main `frontend-engine.stories.tsx`? as it stands now, it has one story only, we can simplify the hierarchy

rgchua: Updated- 

The example suggested by Wei Li should be generally useful in communicating the intention of the story. I suppose one could just use JS to navigate to specific sibling / parent's sibling but putting that all into examples / comments seems overly prescriptive and I think we can agree is ultimately up to the integrating service's needs?

weili-govtech: yes the story and its description is sufficient, may I suggest instead of `console.log()`, use storybook's `action()` method instead? it works like this: https://github.com/LifeSG/web-frontend-engine/blob/main/src/stories/2-frontend-engine/frontend-engine.stories.tsx#L245

weili-govtech (approved):

weili-govtech:
@qroll need your approval too as you have requested changes

qroll (approved):

PR #83, opened 2023-06-12
[BUGFIX][WORKROUND] Temporaly moves rds from peer-deps to deps

v1shva:
**Changes**
[BUGFIX][WORKROUND] Temporaly moves rds from peer-deps to deps so that npm will handle transitive deps for fee correctly

-  delete branch

**Additional information**

-   You may refer to this docs: https://nodejs.org/en/blog/npm/peer-dependencies
- also :  https://blog.npmjs.org/post/618653678433435649/npm-v7-series-arborist-deep-dive.html
- https://stackoverflow.com/questions/18875674/whats-the-difference-between-dependencies-devdependencies-and-peerdependencie

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,10 +50,10 @@
 		"react-hook-form": "^7.34.2",
 		"sanitize-html": "^2.8.1",
 		"use-deep-compare-effect": "^1.8.1",
-		"yup": "^0.32.11"
+		"yup": "^0.32.11",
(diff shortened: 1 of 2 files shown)


package.json
weili-govtech, line 1: - move `@lifesg/react-design-system` below `@js-joda/timezone` so the packages are still sorted alphabetically
- update or regenerate package-lock to keep it in sync

weili-govtech (approved):

PR #86, opened 2023-06-19
Image upload

weili-govtech:
**Changes**
- Fixed SSR error with rendering the image upload component
- Ensure image upload-specific validation rule typings are supported
- delete branch

**Additional information**
image
- Errors were reported when image upload field is included
- It is found to occur in SSR + `validationMode="all"`


src/components/fields/image-upload/image-manager/image-manager.ts
```diff
@@ -137,17 +137,21 @@ export const ImageManager = (props: IProps) => {
...
-					.map(({ dataURL, drawingDataURL, name, uploadResponse }) => ({
-						fileName: name,
-						dataURL: drawingDataURL || dataURL,
-						uploadResponse,
-					})),
-			},
-		});
+		// checking for previous images to prevent firing onChange on first mount
+		// otherwise it will crash on SSR (likely due to calling validation when field is not registered)
+		if (previousImages) {
```
qroll, line 142: this is highly dependent on the initial value of previousImages/images, but I don't see a better way. wondering if `setValue` with `shouldValidate=false` when the field is not dirty would work

weili-govtech: ok just gave it a try and managed to use `setValue` through a combination of `shouldValidate` and `shouldDirty`, agree that this is a better way indeed

qroll (approved):

PR #87, opened 2023-06-20
Fix conditional rendering behaviour when pre-filled

weili-govtech:
**Changes**
- save form values to state in conditional render, this allows default values to be saved and used to decide if the children can be rendered
- delete branch

**Additional information**
- this is to fix a bug when a field is pre-filled, the conditionally rendered fields are not rendered despite fulfilling the conditions


qroll (changes requested):
can a test case be added for this?

weili-govtech:
yup have just added a test for it

qroll (approved):

PR #90, opened 2023-06-21
Update image upload component

Huizz:
… and 

**Changes**
Description of changes...
- delete branch

**Changelog entry**

Image upload component
- Fix colour for image upload description
- Add drop files text
- Align "Add Photos" button the drop files text to right of component

**Additional information**

-   You may refer to figma for the intended design

### DIFF
--- a/src/components/fields/image-upload/image-input/image-input.styles.ts
+++ b/src/components/fields/image-upload/image-input/image-input.styles.ts
@@ -24,10 +24,13 @@ export const Subtitle = styled(Text.Body)`
 `;
 
 export const Content = styled(Text.BodySmall)`
-	color: ${Color.Neutral[3]};
+	color: ${Color.Neutral[4]};
 `;
 
 export const UploadWrapper = styled.div`
+	display: flex;
+	flex-direction: column;
+	align-items: flex-end;
 	margin-top: 2rem;
 `;
 
--- a/src/components/fields/image-upload/image-input/image-input.styles.ts
+++ b/src/components/fields/image-upload/image-input/image-input.styles.ts
@@ -6,7 +6,7 @@ import { Text } from "@lifesg/react-design-system/text";
 import styled from "styled-components";
 
 export interface SubtitleProps {
-	hasDescription?: boolean;
+	$hasDescription?: boolean;
 }
 
 export const Wrapper = styled.div`
@@ -20,7 +20,7 @@ export const Wrapper = styled.div`
 `;
 
 export const Subtitle = styled(Text.Body)<SubtitleProps>`
-	margin-bottom: ${(props) => (props.hasDescription ? "0.5rem" : "1rem")};
+	margin-bottom: ${(props) => (props.$hasDescription ? "0.5rem" : "1rem")};
 `;
 
 export const Content = styled(Text.BodySmall)`
--- a/src/components/fields/image-upload/image-input/image-input.styles.ts
+++ b/src/components/fields/image-upload/image-input/image-input.styles.ts
@@ -5,6 +5,10 @@ import { MediaQuery } from "@lifesg/react-design-system/media";
 import { Text } from "@lifesg/react-design-system/text";
 import styled from "styled-components";
 
+export interface SubtitleProps {
+	hasDescription?: boolean;
+}
+
 export const Wrapper = styled.div`
 	margin: 1.25rem 0;
 	border-radius: 0.25rem;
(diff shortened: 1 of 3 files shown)


weili-govtech:
are the font weights and spacings part of this fix? I notice the title need to have a font-weight of 600 and 0.5rem margin to the description below.

Huizz:
thanks for noticing. 

The font is already loaded as "Open Sans semibold" so it should already be weight 600. not sure why the developer console shows it as 400. 
actual font-weight 400 will look like this
image

Huizz:
as for the spacing. there was this portion in the original styles for `<DragUpload />`
```
	> :not(:last-child) {
		margin-bottom: 1rem;
	}
```

I dont have a reference for when the component does not have the description. would you happen to know if this style is still valid and the 0.5rem only applies if there is a description?

Huizz:
another curious point.

description is optional in `IImageUploadSchema` but required in `IImageInputProps`.
there is no fallback to a default value if description is undefined, before passing it to `<ImageInput />` as part of `IImageInputProps`

should description also be optional  in `IImageInputProps`?

weili-govtech:
> The font is already loaded as "Open Sans semibold" so it should already be weight 600. not sure why the developer console shows it as 400.

whoops my bad, I missed that out, then this is fine.

---

> as for the spacing. there was this portion in the original styles for <DragUpload />

	> :not(:last-child) {
		margin-bottom: 1rem;
	}
> I dont have a reference for when the component does not have the description. would you happen to know if this style is still valid and the 0.5rem only applies if there is a description?

the quoted style for `<DragUpload />` is not meant for the description though. currently there have not been any use case for a component without description but I think we can remove that `0.5rem` if there are no descriptions.

---

> another curious point.

> description is optional in IImageUploadSchema but required in IImageInputProps.
there is no fallback to a default value if description is undefined in IImageUploadSchema, before passing it to <ImageInput /> as part of IImageInputProps

> should description also be optional in IImageInputProps?

yes that's a fair point, can update as optional.

---

note: I removed the figma link as this is a public repo, we can share the mocks through another medium instead.

Huizz:
> the quoted style for <DragUpload /> is not meant for the description though. currently there have not been any use case for a component without description but I think we can remove that 0.5rem if there are no descriptions.

Hmm I styled it in a different way, with the `hasDescription` prop. Do let me know if that is okay

> yes that's a fair point, can update as optional.

Okay this has been updated too

src/components/fields/image-upload/image-input/image-input.styles.ts
```diff
@@ -13,21 +17,21 @@ export const Wrapper = styled.div`
 		/* Generated background-image for the dashed border from https://kovart.github.io/dashed-border-generator/  */
 		return `background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg'%3e%3crect width='100%25' height='100%25' fill='none' rx='4' ry='4' stroke='${color}' stroke-width='4' stroke-dasharray='8%2c 8' stroke-dashoffset='8' stroke-linecap='square'/%3e%3c/svg%3e");`;
 	}}
```
weili-govtech, line 19: why remove the bottom margin for the wrapper? it is meant to space out this field from the next field.

src/components/fields/image-upload/image-input/image-input.tsx
```diff
@@ -6,12 +6,20 @@ import { ImageContext } from "../image-context";
...
+	Subtitle,
+	UploadWrapper,
+	Wrapper,
+} from "./image-input.styles";
 
 interface IImageInputProps extends ISharedImageProps {
 	label: string;
 	buttonLabel?: string | undefined;
-	description: string;
+	description?: string;
```
weili-govtech, line 22:
```suggestion
	description?: string | undefined;
```

weili-govtech (changes requested):

src/components/fields/image-upload/image-input/image-input.styles.ts
Huizz: wrapper only has one child ([file](https://github.com/LifeSG/web-frontend-engine/blob/main/src/components/fields/image-upload/image-input/image-input.tsx#L162)) so I thought it was redundant

Huizz: oh this is not for the last child of Wrapper?

weili-govtech (approved):

PR #99, opened 2023-06-27
Fixes

weili-govtech:
**Changes**
- add rollup postcss to handle leaflet css import
- resolve npm audit warnings
- ensure onChange prop is fired after validation config is initialised
- delete branch


src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
```diff
@@ -140,6 +140,30 @@ describe("frontend-engine", () => {
...
+								validation: [{ min: 5 }],
+							},
+						},
+					},
+				},
+				defaultValues: { [FIELD_ONE_ID]: "a" },
+			}
+		);
+
+		expect(onChange).toBeCalledWith(expect.objectContaining({ [FIELD_ONE_ID]: "a" }), false);
```
qroll, line 164: the assertion doesn't match the test case description, the second argument should be `true` and the default value should have min 5 characters

weili-govtech: oops just fixed it, I updated the tests to cover both positive and negative outcomes

qroll (approved):

PR #109, opened 2023-07-14
Mol 13278

Huizz:
**Changes**
Description of changes...

-  delete branch
- Update default maxAttempts to 3 for getCurrentLocation
- Only call Error.captureStackTrace if Error.captureStackTrace is available

**Changelog entry**

- Fixes on location-input component

### DIFF
--- a/src/components/fields/location-field/location-modal/types.ts
+++ b/src/components/fields/location-field/location-modal/types.ts
@@ -12,7 +12,7 @@ export interface ILocationModalProps
 	className: string;
 	showLocationModal: boolean;
 	formValues?: ILocationFieldValues | undefined;
-	locationModalStyles?: string;
+	locationModalStyles?: string | undefined;
 	onClose: () => void;
 	onConfirm: (values: ILocationFieldValues) => void;
 	updateFormValues: (values: ILocationFieldValues) => void;
--- a/src/components/fields/location-field/location-modal/types.ts
+++ b/src/components/fields/location-field/location-modal/types.ts
@@ -12,6 +12,7 @@ export interface ILocationModalProps
 	className: string;
 	showLocationModal: boolean;
 	formValues?: ILocationFieldValues | undefined;
+	mastheadHeight?: number;
 	onClose: () => void;
 	onConfirm: (values: ILocationFieldValues) => void;
 	updateFormValues: (values: ILocationFieldValues) => void;
--- a/src/components/fields/location-field/location-modal/types.ts
+++ b/src/components/fields/location-field/location-modal/types.ts
@@ -12,7 +12,7 @@ export interface ILocationModalProps
 	className: string;
 	showLocationModal: boolean;
 	formValues?: ILocationFieldValues | undefined;
-	mastheadHeight?: number;
+	locationModalStyles?: string;
 	onClose: () => void;
 	onConfirm: (values: ILocationFieldValues) => void;
 	updateFormValues: (values: ILocationFieldValues) => void;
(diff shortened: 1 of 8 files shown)


src/services/onemap/types.ts
```diff
@@ -60,6 +60,8 @@ export class OneMapError extends Error {
 		super(message);
 		this.name = "OneMapError";
 		this.innerError = error;
-		Error.captureStackTrace(this, OneMapError);
+		if (Error.captureStackTrace) {
+			Error.captureStackTrace(this, OneMapError);
+		}
```
weili-govtech, line 65: can use optional chaining

```suggestion
Error.captureStackTrace?.(this, OneMapError);
```

src/utils/geolocation-helper.ts
```diff
@@ -23,15 +23,14 @@ export namespace GeoLocationHelper {
...
-	 * and whether to disable the error prompt in the LifeSG app (default: false).
+	 * the timeout in milliseconds (default: 2000), the maximum age of a possible cached position in milliseconds (default: 2000)
 	 * A minimum of one attempt will be made, even if maxAttempts is 0.
 	 * @returns A Promise that gets the current location.
 	 * @resolve type: ILocationCoord
 	 * @reject type: GeolocationPositionError | undefined
 	 */
 	export const getCurrentLocation = async (options: IGetCurrentLocationOptions = {}): Promise<ILocationCoord> => {
-		const { maxAttempts = 1, timeout = 2000, maximumAge = 2000 } = options;
+		const { maxAttempts = 3, timeout = 2000, maximumAge = 2000 } = options;
```
weili-govtech, line 33: what's the rationale of changing it? I think the original intention was to offer retry as an option

weili-govtech (changes requested):

src/utils/geolocation-helper.ts
Huizz: oh I asked Hilman and he mentioned the default should be 3 instead so I updated it.

Retry is also not offered as an option now though. so I should be exposing that instead? what's the previous discussion on this?

weili-govtech: ok I have checked with Hilman, it appears it's more ideal to have retry enabled by default. (when maxAttempts=1, it's akin to no retry enabled)

however can we update the comments to reflect the new behaviour
```ts
/**
 * @param options An optional options object that lets you specify the maximum no. of attempts (default: 3),
 * the timeout in milliseconds (default: 2000), the maximum age of a possible cached position in milliseconds (default: 2000),
 * A minimum of one attempt will be made, even if maxAttempts is 0.
 * @returns A Promise that gets the current location.
 * @resolve type: ILocationCoord
 * @reject type: GeolocationPositionError | undefined
*/
```

src/components/fields/location-field/location-modal/types.ts
```diff
@@ -12,6 +12,7 @@ export interface ILocationModalProps
 	className: string;
 	showLocationModal: boolean;
 	formValues?: ILocationFieldValues | undefined;
+	locationModalStyles?: string;
```
weili-govtech, line 15:
```suggestion
	locationModalStyles?: string | undefined;
```

src/components/fields/location-field/types.ts
```diff
@@ -14,6 +14,7 @@ export interface ILocationFieldSchema<V = undefined>
 		Pick<ILocationInputProps, "locationInputPlaceholder">,
 		Pick<IStaticMapProps, "staticMapPinColor"> {
 	className?: string;
+	locationModalStyles?: string;
```
weili-govtech, line 17:
```suggestion
	locationModalStyles?: string | undefined;
```

weili-govtech (changes requested):

weili-govtech (approved):

PR #119, opened 2023-07-24
MOL-13392

Huizz:
**Changes**
Return pin location if reverse geocode returns empty array

-  delete branch

**Changelog entry**

-   Update location search to return pin location if reverse geocode endpoint returns empty array

### DIFF
--- a/src/components/fields/location-field/location-helper.ts
+++ b/src/components/fields/location-field/location-helper.ts
@@ -156,6 +156,22 @@ export namespace LocationHelper {
 	// =========================================================================
 	export const reverseGeocode = async ({ options, ...others }: TReverseGeocodeParams): Promise<IResultsMetaData> => {
 		const locationList = await mapService.reverseGeocode(others);
+
+		if (locationList.length === 0) {
+			const lat = others.latitude || 0;
+			const lng = others.longitude || 0;
+
+			return {
+				results: [
+					{
+						address: `${Math.round(lat * 100) / 100}, ${Math.round(lng * 100) / 100}`,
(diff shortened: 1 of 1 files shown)


src/components/fields/location-field/location-helper.ts
```diff
@@ -156,6 +160,22 @@ export namespace LocationHelper {
 	// =========================================================================
 	export const reverseGeocode = async ({ options, ...others }: TReverseGeocodeParams): Promise<IResultsMetaData> => {
 		const locationList = await mapService.reverseGeocode(others);
```
weili-govtech, line 162: can we have the `isCoordinateInBounds()` check here instead?

because this function is also manually filtering `JOHOR (MALAYSIA)` and supports an `excludeNonSG` option, we can have it throw an error instead.

in addition, `reverseGeocode()` is used in other places but don't have the same boundary check, so it should have that check in here to maintain the same behaviour.

src/components/fields/location-field/location-helper.ts
```diff
@@ -155,7 +159,26 @@ export namespace LocationHelper {
...
+			return {
+				results: [
+					{
+						address: `Pin location ${Math.round(lat * 100) / 100}, ${Math.round(lng * 100) / 100}`,
+						lat,
+						lng,
+					},
+				],
+			};
+		}
```
weili-govtech, line 180: can remove since there is a similar logic at the end of the function

weili-govtech (approved):

Huizz:
Closing this to open a new PR

PR #124, opened 2023-07-27
[MOL-13378][MA] add test for fetchLocationList helper function

mauw-paloit:
**Changes**

- Add new test file for location-helper
- Add test for fetchLocationList helper function error handling

### DIFF
--- a/src/__tests__/utils/location-helper.spec.ts
+++ b/src/__tests__/utils/location-helper.spec.ts
@@ -4,7 +4,7 @@ import { OneMapService } from "../../services";
 import { LocationHelper } from "../../components/fields/location-field/location-helper";
 
 describe("when reverseGeoCode returns an error", () => {
-	let mockError;
+	let mockError: Error;
 	const mockOnError = jest.fn();
 	const mockRef: React.MutableRefObject<AbortController | null> = {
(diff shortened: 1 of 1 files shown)


src/__tests__/utils/location-helper.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+describe("when reverseGeoCode returns an error", () => {
+	let mockError;
+	const mockOnError = jest.fn();
+	const mockRef: React.MutableRefObject<AbortController | null> = {
+		current: null,
+	};
+
+	beforeEach(() => {
+		jest.resetAllMocks();
```
weili-govtech, line 14: as your assertions are nested within a try... catch block, it may not run if your function resolves successfully.
you can add `expect.hasAssertions();` to make sure each test has some assertions.

weili-govtech (changes requested):

src/__tests__/utils/location-helper.spec.ts
mauw-paloit: Thanks, I've added this in each of the test cases

weili-govtech (approved):

PR #127, opened 2023-07-27
Mol 13392 fix

Huizz:
**Changes**
Description of changes...

- Fix issue when default location is a pin location and user opens the map modal again
- Delete branch

**Changelog entry**

-    Fix issue when default location is a pin location and user opens the map modal again


src/components/fields/location-field/location-modal/location-search/location-search.tsx
weili-govtech, line 1: can we also extract the lat and lng values out of `Pin location {lat}, {lng}` and do reverse geocoding based on it?

weili-govtech (approved):

PR #130, opened 2023-08-04
Populate conditionally rendered fields

qroll:
**Changes**

- Track user's historical field inputs in a `FormValues` context
- Update fields on changes (via `watch`) and reset
- Introduce a `restoreMode` that configures how conditionally rendered fields are populated when they are toggled
- Added and reorganised stories under `Condtional Rendering/Rules` and `Condtional Rendering/Restoring Values`
- [delete] branch

### DIFF
--- a/src/components/fields/chips/chips.tsx
+++ b/src/components/fields/chips/chips.tsx
@@ -28,12 +28,19 @@ export const Chips = (props: IGenericFieldProps<IChipsSchema>) => {
 	const [stateValue, setStateValue] = useState<string[]>(value || []);
 	const [showTextarea, setShowTextarea] = useState(false);
 	const [multi, setMulti] = useState(true);
-	const { setValue } = useFormContext();
+	const { setValue, unregister } = useFormContext();
 	const { setFieldValidationConfig, removeFieldValidationConfig } = useValidationConfig();
 
 	// =============================================================================
 	// EFFECTS
 	// =============================================================================
+	useEffect(() => {
+		if (!showTextarea) {
+			// at the start, the textarea's default value has to be manually cleared from the form state
+			unregister(getTextareaId());
+		}
+	}, []);
+
 	useEffect(() => {
 		const isRequiredRule = validation?.find((rule) => "required" in rule);
 		const maxRule = validation?.find((rule) => "max" in rule);
--- a/src/components/fields/chips/chips.tsx
+++ b/src/components/fields/chips/chips.tsx
@@ -92,7 +92,6 @@ export const Chips = (props: IGenericFieldProps<IChipsSchema>) => {
 		setShowTextarea(show);
 		if (!show) {
 			removeFieldValidationConfig(getTextareaId());
-			setValue(getTextareaId(), undefined);
 		}
 	};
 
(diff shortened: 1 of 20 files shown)


src/components/frontend-engine/types.ts
```diff
@@ -106,6 +107,7 @@ export type TFrontendEngineFieldSchema<V = undefined> =
 export type TFrontendEngineValues<T = any> = Record<keyof T, T[keyof T]>;
 export type TRevalidationMode = Exclude<keyof ValidationMode, "onTouched" | "all">;
 export type TValidationMode = keyof ValidationMode;
+export type TRestoreMode = "default-value" | "user-input";
```
weili-govtech, line 110: apologies, I missed this out, can we have another mode to not restore the values, this is to support our current behaviour of no restoring the values at all

src/components/frontend-engine/types.ts
```diff
@@ -74,6 +74,7 @@ export interface IFrontendEngineData<V = undefined> {
 	validationMode?: TValidationMode | undefined;
 	/** Additional properties to mutate the sections schema on-the-fly */
 	overrides?: Record<string, RecursivePartial<TFrontendEngineFieldSchema<V>>> | undefined;
+	restoreMode?: TRestoreMode | undefined;
```
weili-govtech, line 77: can add jsdocs comment to explain what is `restoreMode`

src/components/frontend-engine/form-values/context-provider.tsx
```diff
@@ -0,0 +1,28 @@
+import { Dispatch, MutableRefObject, ReactElement, SetStateAction, createContext, useRef, useState } from "react";
+
```
weili-govtech, line 2: add simple explanation about the need to use form values context

src/components/fields/chips/chips.tsx
```diff
@@ -27,12 +28,19 @@ export const Chips = (props: IGenericFieldProps<IChipsSchema>) => {
...
+	const { setValue, unregister } = useFormContext();
 	const { setFieldValidationConfig, removeFieldValidationConfig } = useValidationConfig();
 
 	// =============================================================================
 	// EFFECTS
 	// =============================================================================
+	useEffect(() => {
+		if (!showTextarea) {
+			// at the start, the textarea's default value has to be manually cleared from the form state
+			unregister(getTextareaId());
```
weili-govtech, line 40: just to clarify, is this a temp fix before you make a complete fix in another PR?
I wonder if `unregister` can be used in FieldWrapper or ConditionalRenderer

weili-govtech (changes requested):
do pull the latest commits from master because there is an update to the filter clear button that is essentially resetting the fields under it

src/components/fields/chips/chips.tsx
qroll: yup this is temporary. I'm intending to run `unregister` for each `idsToDelete` in `ConditionalRenderer`. `FieldWrapper` won't work for all cases as it requires the component to unmount which doesn't happen on fields nested within conditionally rendered fields

src/components/frontend-engine/types.ts
qroll: added a `"none"` mode which clears the values similarly to the reset button

weili-govtech (approved):

PR #133, opened 2023-08-07
MOL-13460 Bug fix: Pin location not in result list on edit location

shengyang-c-dcube:
**Changes**
- Made use of the `address` property in `formValues` to check if the pin location value is used.
- Added `useFallback` arg in `fetchAddress` function, to utilize the pin location value without invoking the `searchByAddress` and `formatAddressFromGeocodeInfo`.
- Populate the result list with the previously selected pin location on revisits to the location modal.
- Delete branch.

**Additional information**
-   Ticket [MOL-13460](https://jira.ship.gov.sg/browse/MOL-13460)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@lifesg/web-frontend-engine",
-	"version": "1.0.0-alpha.15",
+	"version": "1.0.0-alpha.16",
 	"description": "A frontend engine for LifeSG React apps",
 	"main": "dist/cjs/index.js",
 	"module": "dist/index.js",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,6 @@
 	"typings": "dist/index.d.ts",
 	"scripts": {
 		"build": "npm run rollup && npm run post:build",
-		"build:ci": "./scripts/build.sh",
 		"build-watch": "rollup -c -w",
 		"build-storybook": "build-storybook",
 		"build-check": "npm run rollup",
--- a/package.json
+++ b/package.json
@@ -1,12 +1,13 @@
 {
 	"name": "@lifesg/web-frontend-engine",
-	"version": "1.0.0-alpha.16",
+	"version": "1.0.0-alpha.15",
 	"description": "A frontend engine for LifeSG React apps",
 	"main": "dist/cjs/index.js",
 	"module": "dist/index.js",
 	"typings": "dist/index.d.ts",
 	"scripts": {
 		"build": "npm run rollup && npm run post:build",
+		"build:ci": "./scripts/build.sh",
 		"build-watch": "rollup -c -w",
 		"build-storybook": "build-storybook",
 		"build-check": "npm run rollup",
--- a/src/components/fields/location-field/location-helper.ts
+++ b/src/components/fields/location-field/location-helper.ts
@@ -143,22 +143,18 @@ export namespace LocationHelper {
 		}
 	};
 
-	export const checkAndSetPinLocationAsResult = (query: string, lat: number, lng: number): IResultsMetaData => {
-		// Extract out the lat and lng values from the query.
-		const [queryLat, queryLng] = query
+	export const checkAndSetPinLocationAsResult = (query: string): IResultsMetaData => {
+		const [lat, lng] = query
 			.split(":")[1]
 			.split(",")
 			.map((value) => parseFloat(value.trim()));
 		let parsedResult = [];
-		// Compare the lat, lng values of the query with the lat, lng values of the formValues.
-		// This is done to prevent the user from filling in their own lat, lng values in the form field.
-		if (lat && lng && lat === queryLat && lng === queryLng) {
-			// If the lat, lng values tally, we produce the result item to be put into the result list.
+		if (LocationHelper.isCoordinateInBounds({ lat: lat, lng: lng })) {
 			parsedResult = [
 				{
 					address: query,
-					lat: queryLat,
-					lng: queryLng,
+					lat: lat,
+					lng: lng,
 				},
 			];
 		}
--- a/src/components/fields/location-field/location-helper.ts
+++ b/src/components/fields/location-field/location-helper.ts
@@ -107,71 +107,67 @@ export namespace LocationHelper {
 	export const fetchAddress = async (
 		query: string,
 		pageNumber: number,
-		lat: number,
-		lng: number,
 		onSuccess?: (results: IResultsMetaData) => void,
 		onFail?: (error: unknown) => void
 	) => {
 		if (!query) return;
 		try {
-			// if query contains "Pin location" i.e. in address field.
-			const regex = /^pin location/;
-			if (regex.test(query.toLowerCase())) {
-				// Extract out the lat and lng values from the query.
-				const [queryLat, queryLng] = query
-					.split(":")[1]
-					.split(",")
-					.map((value) => parseFloat(value.trim()));
-				let parsedResult = [];
-				// Compare the lat, lng values of the query with the lat, lng values of the formValues.
-				// This is done to prevent the user from filling in their own lat, lng values in the form field.
-				if (lat && lng && lat === queryLat && lng === queryLng) {
-					// If the lat, lng values tally, we produce the result item to be put into the result list.
-					parsedResult = [
-						{
-							address: query,
-							lat: queryLat,
-							lng: queryLng,
-						},
-					];
-				}
-				onSuccess?.({
-					results: parsedResult,
-					apiPageNum: pageNumber,
-					totalNumPages: pageNumber,
-				});
-			} else {
-				const { results, pageNum, totalNumPages } = await searchByAddress({
-					searchValue: query,
-					getAddressDetails: OneMapBoolean.YES,
-					returnGeom: OneMapBoolean.YES,
-					pageNum: pageNumber,
-				});
+			const { results, pageNum, totalNumPages } = await searchByAddress({
+				searchValue: query,
+				getAddressDetails: OneMapBoolean.YES,
+				returnGeom: OneMapBoolean.YES,
+				pageNum: pageNumber,
+			});
 
-				const parsedResults = results.map((obj) => {
-					const address = formatAddressFromGeocodeInfo(obj, true);
-					return {
-						address: address,
-						blockNo: obj.BLK_NO,
-						building: obj.BUILDING,
-						postalCode: obj.POSTAL,
-						roadName: obj.ROAD_NAME,
-						lat: parseFloat(obj.LATITUDE) || undefined,
-						lng: parseFloat(obj.LONGITUDE) || undefined,
-						x: parseFloat(obj.X) || undefined,
-						y: parseFloat(obj.Y) || undefined,
-					};
-				});
+			const parsedResults = results.map((obj) => {
+				const address = formatAddressFromGeocodeInfo(obj, true);
+				return {
+					address: address,
+					blockNo: obj.BLK_NO,
+					building: obj.BUILDING,
+					postalCode: obj.POSTAL,
+					roadName: obj.ROAD_NAME,
+					lat: parseFloat(obj.LATITUDE) || undefined,
+					lng: parseFloat(obj.LONGITUDE) || undefined,
+					x: parseFloat(obj.X) || undefined,
+					y: parseFloat(obj.Y) || undefined,
+				};
+			});
 
-				onSuccess?.({
-					results: parsedResults,
-					apiPageNum: pageNum,
-					totalNumPages,
-				});
-			}
+			onSuccess?.({
+				results: parsedResults,
+				apiPageNum: pageNum,
+				totalNumPages,
+			});
 		} catch (error) {
 			onFail?.(new OneMapError(error));
 		}
+	};
+
+	export const checkAndSetPinLocationAsResult = (query: string, lat: number, lng: number): IResultsMetaData => {
+		// Extract out the lat and lng values from the query.
+		const [queryLat, queryLng] = query
+			.split(":")[1]
+			.split(",")
+			.map((value) => parseFloat(value.trim()));
+		let parsedResult = [];
+		// Compare the lat, lng values of the query with the lat, lng values of the formValues.
+		// This is done to prevent the user from filling in their own lat, lng values in the form field.
+		if (lat && lng && lat === queryLat && lng === queryLng) {
+			// If the lat, lng values tally, we produce the result item to be put into the result list.
+			parsedResult = [
+				{
+					address: query,
+					lat: queryLat,
+					lng: queryLng,
+				},
+			];
+		}
+		return {
+			results: parsedResult,
+			apiPageNum: 1,
+			totalNumPages: 1,
+		};
 	};
 
 	export const fetchSingleLocationByAddress = async (
@@ -182,8 +178,6 @@ export namespace LocationHelper {
 		await debounceFetchAddress(
 			address,
 			1,
-			undefined,
-			undefined,
 			(res) => {
 				onSuccess(res.results?.[0] || undefined);
 			},
@@ -273,6 +267,12 @@ export namespace LocationHelper {
 	export const hasGotAddressValue = (value?: string): boolean => {
 		const lowercased = value?.toLowerCase();
 		return !!value && lowercased !== "nil" && lowercased !== "null";
+	};
+
+	export const hasGotPinLocationValue = (value?: string): boolean => {
+		if (!value) return false;
+		const regex = /^Pin location:\s*-?\d+(\.\d+)?,\s*-?\d+(\.\d+)?$/;
+		return regex.test(value);
 	};
 
 	export const formatAddressFromGeocodeInfo = (
--- a/src/components/fields/location-field/location-helper.ts
+++ b/src/components/fields/location-field/location-helper.ts
@@ -103,7 +103,7 @@ export namespace LocationHelper {
 		}
 	};
 
-	// Added useFallback arg to handle pin location address.
+	// Added lat, lng arguments to handle pin location address in query field.
 	export const fetchAddress = async (
 		query: string,
 		pageNumber: number,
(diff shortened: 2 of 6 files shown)


src/components/fields/location-field/location-modal/location-modal.tsx
```diff
@@ -97,6 +97,12 @@ const LocationModal = ({
 	useEffect(() => {
 		if (!window) return;
 
+		// If formValues persist upon re-entry into the location modal,
+		// then set selectedAddressInfo accordingly.
+		if (formValues) {
+			setSelectedAddressInfo(formValues);
+		}
+
```
weili-govtech, line 105: what's the rationale of setting the formValues when it will be set at: https://github.com/LifeSG/web-frontend-engine/blob/MOL-13392-fix/src/components/fields/location-field/location-modal/location-search/location-search.tsx#L171

weili-govtech (changes requested):

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@lifesg/web-frontend-engine",
-	"version": "1.0.0-alpha.15",
+	"version": "1.0.0-alpha.16",
```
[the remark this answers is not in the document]
shengyang-c-dcube: Noted. Will revert this. Sorry about it.

src/components/fields/location-field/location-helper.ts
```diff
@@ -103,41 +103,60 @@ export namespace LocationHelper {
 		}
 	};
 
+	// Added useFallback arg to handle pin location address.
 	export const fetchAddress = async (
 		query: string,
 		pageNumber: number,
+		useFallback: boolean,
```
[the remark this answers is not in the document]
Huizz: + 1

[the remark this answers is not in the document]
shengyang-c-dcube: Since `formatAddressFromGeocodeInfo` takes in an arg `fallbackToLatLng` that would produce the pin location as the `address`, I was thinking along the lines of declaring `useFallback` to inform the `fetchAddress` function of the `query` that it is handling.

Now, in retrospect, might have been better to handle it in a different function that returns the result list.

src/components/fields/location-field/location-helper.ts
```diff
@@ -103,41 +103,67 @@ export namespace LocationHelper {
 		}
 	};
 
+	// Added useFallback arg to handle pin location address.
```
Huizz, line 106: redundant comment after the updates

src/components/fields/location-field/location-helper.ts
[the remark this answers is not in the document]
Huizz: the parameter for `formatAddressFromGeocodeInfo` is also quite redundant. there are no calls that do not pass in true as the value for `fallbackToLatLng`

if you traced through the code, you can see that the previous change was introduced here
```
if (parsedLocationList.length === 0) {
	const address = `Pin location: ${lat}, ${lng}`;
	return {
		results: [
			{
				address,
				lat,
				lng,
				displayAddressText: address,
			},
		],
	};
}
```
to be able to let user select a pin location

it does not rely on `formatAddressFromGeocodeInfo` as there are no results in `locationList`

src/components/fields/location-field/location-helper.ts
```diff
@@ -103,41 +103,67 @@ export namespace LocationHelper {
...
-				pageNum: pageNumber,
-			});
+			const regex = /^pin location/;
+			if (regex.test(query.toLowerCase())) {
+				const [queryLat, queryLng] = query
+					.split(":")[1]
+					.split(",")
+					.map((value) => parseFloat(value.trim()));
+				let parsedResult = [];
+				if (lat && lng && lat === queryLat && lng === queryLng) {
```
Huizz, line 124: would there ever be a scenario that the formvalue.lat and formavalue.lng differs from the values taken from query?

asking bcos I'm wondering why we need to pass in lat lng to `fetchAddress`

src/components/fields/location-field/location-modal/location-search/location-search.tsx
```diff
@@ -320,8 +329,13 @@ export const LocationSearch = ({
 	};
 
 	const handleClearInput = () => {
+		// Only set result state as "pristine" when there is no valid formValues available.
```
Huizz, line 332: curious, why so?

src/components/fields/location-field/location-helper.ts
shengyang-c-dcube: I will remove it. Thanks.

src/components/fields/location-field/location-helper.ts
shengyang-c-dcube: It was done with the intention of comparing the `lat` and `lng` values in the `query` (for "Pin location: ...") with the `lat` and `lng` values from `formValue`.

In my view, it only impacts queries that contain "Pin location" in its value.

src/components/fields/location-field/location-helper.ts
```diff
@@ -103,41 +103,67 @@ export namespace LocationHelper {
 		}
 	};
 
+	// Added useFallback arg to handle pin location address.
 	export const fetchAddress = async (
 		query: string,
 		pageNumber: number,
+		lat: number,
```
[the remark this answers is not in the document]
shengyang-c-dcube: Yeah, there is the possibility of it being undefined.

src/components/fields/location-field/location-modal/location-search/location-search.tsx
shengyang-c-dcube: Upon revisiting the location modal multiple times, after initially setting a location, without confirming the location selection in subsequent visits, the result state gets set to "pristine", so the result list is cleared.

If the values in `formValues` exists, meaning that the help-neighbour FE is persisting those values, then we do not set the result state to "pristine".

src/components/fields/location-field/location-modal/location-modal.tsx
shengyang-c-dcube: Yes, you are right. I have removed this section of code.

src/components/fields/location-field/location-modal/location-search/location-search.tsx
```diff
@@ -247,6 +251,7 @@ export const LocationSearch = ({
 		debounceFetchAddress(
 			parsedString,
 			1,
+			formValues?.address !== undefined && formValues?.address === parsedString ? true : false,
```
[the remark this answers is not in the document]
shengyang-c-dcube: I agree that this is neater.

However, due to the code change to have it rely on formValues' `lat` and `lng`, we are not using the `useFallback` boolean arg.

src/components/fields/location-field/location-modal/location-search/location-search.tsx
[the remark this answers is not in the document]
shengyang-c-dcube: Updated.

src/components/fields/location-field/location-helper.ts
shengyang-c-dcube: There can be a scenario where the `lat`, `lng` values taken from `formValues` differ from the values taken from `query`.

In such cases, the result list will not be populated with the result of a user-modified "Pin location: lat, lng" string, but will fallback on the usual location search flow, which would produce the "No results found..." text.

src/components/fields/location-field/location-helper.ts
```diff
@@ -151,6 +170,7 @@ export namespace LocationHelper {
 		await debounceFetchAddress(
 			address,
 			1,
+			false,
```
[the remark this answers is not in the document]
shengyang-c-dcube: I had to add that arg in as `fetchAddress` requires the 3rd arg `useFallback` inside.

This has been changed to use `lat` and `lng` from `formValues` as the new arguments in the latest commit(s).

package.json
```diff
@@ -7,6 +7,7 @@
 	"typings": "dist/index.d.ts",
 	"scripts": {
 		"build": "npm run rollup && npm run post:build",
+		"build:ci": "./scripts/build.sh",
```
weili-govtech, line 10: remove extraneous script

src/components/fields/location-field/location-helper.ts
weili-govtech: may I know what are the scenarios where the latlng is different from the query? just trying to understand how are we to deal with such scenarios. with this logic, it will yield no results and the user will be confused. I am under the impression we can search based on the query provided?

in addition, for the checking of the query, instead of checking for the presence of `pin location`, manually lower casing it to test and making undefined checks, may I suggest to use the same regex from location-search for it:
`^(pin location:) -?\d{0,3}.\d*, -?\d{0,3}.\d*$`

- this will help check for the presence of `pin location`, exact query format and the presence of the lat and lng
- there won't be the need to check if `lat` or `lng` is undefined; `queryLat` and `queryLng` is definitely defined with the regex check, so the equality checks will fail if `lat` or `lng` is undefined

src/components/fields/location-field/location-helper.ts
```diff
@@ -103,41 +103,72 @@ export namespace LocationHelper {
 		}
 	};
 
+	// Added lat, lng arguments to handle pin location address in query field.
```
weili-govtech, line 106: unnecessary comment, there is no need to explain what has been changed

weili-govtech (changes requested):

src/components/fields/location-field/location-helper.ts
shengyang-c-dcube: _may I know what are the scenarios where the latlng is different from the query? just trying to understand how are we to deal with such scenarios._

If the user attempts to input/alter their own lat, lng values and "Pin location" text in the search field, then there is a possibility for it to be different from the lat, lng values introduced into the formValues.

package.json
shengyang-c-dcube: Removing it. Sorry about the inclusion.

src/components/fields/location-field/location-helper.ts
Huizz: how do we end up in this scenario?

> There can be a scenario where the lat, lng values taken from formValues differ from the values taken from query.

src/components/fields/location-field/location-modal/location-search/location-search.tsx
Huizz: hmm open to discussion but I would think it's better to handle it this way (and this is what I understand how the component handles it currently)

- result list is re-built every time the modal is opened, based on existing value in the form
- so there's no issue to clear or destroy this result list when location modal closes
- if user clicks on the map or searches for another location while the modal is open, the list is also updated
- so no issues to clear the previous result list here

Huizz: maybe @weili-govtech could give more insights into this? or the change made by Derek is okay?

src/components/fields/location-field/location-helper.ts
shengyang-c-dcube: Upon revisiting the location modal after navigating back from review screen, we have the `formValues.address`, `formValues.lat`, `formValues.lng` as well as the `queryString` from the prior confirmation.

The scenario where the values can differ is when the user attempts to alter the `queryString` to produce a different pin location value.

src/components/fields/location-field/location-modal/location-search/location-search.tsx
shengyang-c-dcube: You can try replicating this other bug I discovered by:
Step 1: Open location modal, select a location, confirm and close modal.
Step 2: Re-open location modal, see that the location is still selected, do not confirm, just close modal.
Step 3: Re-open location modal, prior confirmed location is in the `queryString`, but the location is no longer "confirmed" within the modal. 

Closing the modal thereafter with no selection or confirmation of a new location will have no impact on the existing form values. However, if this is the intended behaviour, then I will remove my changes. I just thought that it might be confusing to the user.

src/components/fields/location-field/location-helper.ts
Huizz: oh is the user able to type in a value into the input? afaik it needs to be set via code

shengyang-c-dcube: I meant the string value for searching location. I did not want the users to circumvent the search by typing their own "Pin location: lat, lng", that should produce no results.

Huizz: hmm I thought the user is not able to type in their own Pin location: lat, lng in the first place, that's why was thinking we don't even need to prevent the scenario bcos it doesn't happen.

But I may be wrong here. the user can type ?

Huizz: hmm if I try something like this it won't search for me. tested on dev env
image

src/components/fields/location-field/location-modal/location-search/location-search.tsx
weili-govtech: > hmm open to discussion but I would think it's better to handle it this way (and this is what I understand how the component handles it currently)
> 
> * result list is re-built every time the modal is opened, based on existing value in the form
> * so there's no issue to clear or destroy this result list when location modal closes
> * if user clicks on the map or searches for another location while the modal is open, the list is also updated
> * so no issues to clear the previous result list here

yup the understanding is correct, basically the result list will refresh every time the modal is opened, doesn't matter what happened before that, it will always query again based on the form values

> You can try replicating this other bug I discovered by: Step 1: Open location modal, select a location, confirm and close modal. Step 2: Re-open location modal, see that the location is still selected, do not confirm, just close modal. Step 3: Re-open location modal, prior confirmed location is in the `queryString`, but the "Confirm" button is no longer enabled within the modal.
> 
> Closing the modal thereafter with no selection or confirmation of a new location will have no impact on the existing form values. However, if this is the intended behaviour, then I will remove my changes. I just thought that it might be confusing to the user.

the disabling of confirm button looks like a bug, if there is a confirmed location, every time user brings up the modal, the result list needs to be populated based on the form values with the confirmed location selected in the address list and the search bar populated with the confirmed location address.

shengyang-c-dcube: Noted. Will revert the changes made, pertaining to the setting of the "pristine" state, and look into the button bug.

Huizz: @shengyang-c-dcube 

To sum up the conversation here:
1. The behaviour you reported is a bug
2. The actual behaviour should be that the list is refetched every time the modal is opened
3. The solution implemented in this PR patches the bug but does not resolve the underlying issue nor is it the intended behaviour

We'll report this as a separate bug, not to be fixed as part of your ticket.

@weili-govtech do you guys track bugs right now? or this has to be on the SH board? if it's on the SH board it may have to go to backlog for now.

weili-govtech: we track the bugs on a separate kanban board, you can leave it in the SH backlog if it is not critical. however can you put these tickets under `Web Frontend Engine` (MOL-11408) epic?

weili-govtech: apologies, instead of tagging with the epic, can you raise a bug ticket in this repo with a link to the jira ticket?

src/components/fields/location-field/location-helper.ts
Huizz: clarified this over code review.

1. The above screenshot is a result of the original bug that this PR is trying to fix
2. We will let the user search for a lat lng using `Pin location: <lat>, <lng>` as long as it's within the map bounds.

@weili-govtech what do you think about point 2?

shengyang-c-dcube:
Changes made:
- Allowed the input of pin location by the user.
- Checks made to see if lat, lng is within boundary.

src/components/fields/location-field/location-helper.ts
```diff
@@ -143,6 +143,28 @@ export namespace LocationHelper {
...
+				},
+			];
+		}
+		return {
+			results: parsedResult,
+			apiPageNum: 1,
+			totalNumPages: 1,
+		};
+	};
+
```
shengyang-c-dcube, line 167: In the case where the coordinates provided by the user is out of bounds, the behaviour is no pin location "search result" is produced in the list, so the user cannot confirm the change.

src/components/fields/location-field/location-helper.ts
weili-govtech: I am fine with point 2 because it is cleaner and behaves more consistently. 
however there is a need to confirm with BA and UX because OS does not behave the same, user cannot search in this way.

src/components/fields/location-field/location-modal/location-search/location-search.tsx
weili-govtech, line 1: now that there is the `hasGotPinLocationValue()` helper function, you can implement the same check on line 207

src/components/fields/location-field/location-helper.ts
```diff
@@ -143,6 +143,28 @@ export namespace LocationHelper {
...
+		let parsedResult = [];
+		if (LocationHelper.isCoordinateInBounds({ lat: lat, lng: lng })) {
+			parsedResult = [
+				{
+					address: query,
+					lat: lat,
+					lng: lng,
+				},
+			];
+		}
```
weili-govtech, line 160: just suggesting some shorthands to use
```suggestion
		if (LocationHelper.isCoordinateInBounds({ lat, lng })) {
			parsedResult = [
				{
					address: query,
					lat,
					lng,
				},
			];
		}
```

Huizz (approved):

src/components/fields/location-field/location-helper.ts
```diff
@@ -143,6 +143,28 @@ export namespace LocationHelper {
 		}
 	};
 
+	export const checkAndSetPinLocationAsResult = (query: string): IResultsMetaData => {
+		const [lat, lng] = query
+			.split(":")[1]
+			.split(",")
+			.map((value) => parseFloat(value.trim()));
```
weili-govtech, line 150: no need to trim before parseFloat

this function only work with a specific string format otherwise it will crash. it'll be a good idea to check the format via `hasGotPinLocationValue()` before running it

weili-govtech (changes requested):

weili-govtech (approved):

PR #135, opened 2023-08-10
[RBS-427][VM] Add date range component

v1shva:
**Changes**
- Add date range component

-   delete branch

**Additional information**

-   You may refer to this [RBS-427](https://jira.ship.gov.sg/browse/RBS-427)

### DIFF
--- a/src/components/fields/date-range-field/date-range-field.tsx
+++ b/src/components/fields/date-range-field/date-range-field.tsx
@@ -8,7 +8,7 @@ import { DateTimeHelper, TestHelper } from "../../../utils";
 import { useValidationConfig } from "../../../utils/hooks";
 import { IGenericFieldProps } from "../../frontend-engine/types";
 import { ERROR_MESSAGES } from "../../shared";
-import { IDateRangeFieldSchema } from "./types";
+import { IDateRangeFieldSchema, TDateRangeInputType } from "./types";
 
 const DEFAULT_DATE_FORMAT = "uuuu-MM-dd";
 const DEFAULT_DATE_FORMATTER = DateTimeFormatter.ofPattern(DEFAULT_DATE_FORMAT)
@@ -43,7 +43,6 @@ export const DateRangeField = (props: IGenericFieldProps<IDateRangeFieldSchema>)
 		const notPastRule = validation?.find((rule) => "notPast" in rule);
 		const minDateRule = validation?.find((rule) => "minDate" in rule);
 		const maxDateRule = validation?.find((rule) => "maxDate" in rule);
-		const excludedDatesRule = validation?.find((rule) => "excludedDates" in rule);
 		const isRequiredRule = validation?.find((rule) => "required" in rule);
 
 		const minDate = DateTimeHelper.toLocalDateOrTime(minDateRule?.["minDate"], dateFormat, "date");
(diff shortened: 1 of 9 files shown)


src/components/fields/date-range-field/date-range-field.tsx
```diff
@@ -0,0 +1,272 @@
...
+						const localDateFrom = DateTimeHelper.toLocalDateOrTime(value.from, dateFormat, "date");
+						const localDateTo = DateTimeHelper.toLocalDateOrTime(value.to, dateFormat, "date");
+						try {
+							const mappedexcludedDates = excludedDatesRule["excludedDates"].map((date) =>
+								DateTimeHelper.toLocalDateOrTime(date, dateFormat, "date")
+							);
+							for (const excludedDate of mappedexcludedDates) {
+								if (localDateFrom.isBefore(excludedDate) && localDateTo.isAfter(excludedDate))
+									return false;
+							}
```
qroll, line 134: I don't have context or access to the ticket, so I'm not sure if the ACs cover this. clarifying if the following are expected behaviour:

1. for range variant, disabled date within (exclusive) start/end date - ALLOWED
2. for range variant, disabled date as start/end date - ?
3. for week variant, disabled date can be the start/end date as long as at least one other day within the week is selectable - ?
4. i.e. for week variant, a fully disabled week cannot be selected - ?

src/components/fields/date-range-field/types.ts
```diff
@@ -0,0 +1,20 @@
...
+	minDate?: string | undefined;
+	maxDate?: string | undefined;
+	excludedDates?: string[] | undefined;
+}
+
+export interface IDateRangeFieldSchema<V = undefined>
+	extends IFrontendEngineBaseFieldJsonSchema<"date-range-field", V, IDateRangeFieldValidationRule>,
+		TComponentOmitProps<DateRangeInputProps, "valueEnd"> {
+	dateFormat?: string | undefined;
+	variant?: Exclude<Variant, "single"> | undefined;
```
qroll, line 19: I believe `DateRangeInputProps` already has a variant so there's no need to redefine it

src/components/fields/date-range-field/date-range-field.tsx
weili-govtech: +1
why are some of the validation rules skipped for the `week` variant?

src/components/fields/date-range-field/date-range-field.tsx
```diff
@@ -0,0 +1,259 @@
...
+					}
+				)
+				.test("is-date", ERROR_MESSAGES.DATE_RANGE.INVALID, (value) => {
+					if (!value || value.from === "" || value.to === "") return true;
+					if (value.from === undefined || value.to === undefined) return true;
+					if (!isValidDate(value.from) || !isValidDate(value.to)) return false;
+					return (
+						!!DateTimeHelper.toLocalDateOrTime(value.from, dateFormat, "date") ||
+						!!DateTimeHelper.toLocalDateOrTime(value.to, dateFormat, "date")
+					);
```
weili-govtech, line 74: this will return truthy if either value.from or value.to is valid but both needs to be valid

weili-govtech, line 74: can simplify the checks by using lodash's `isEmpty` and combining the return logic

```
if (isEmpty(value?.from) || isEmpty(value?.to)) return true;
return (
	isValidDate(value.from) &&
	isValidDate(value.to) &&
	!!DateTimeHelper.toLocalDateOrTime(value.from, dateFormat, "date") &&
	!!DateTimeHelper.toLocalDateOrTime(value.to, dateFormat, "date")
);
```

src/stories/3-fields/date-range-field/date-range-field.stories.tsx
weili-govtech, line 1: we have just switched to storybook 7, there is a need to change some of the typings, you can refer to how other stories are written

weili-govtech (changes requested):

src/components/fields/date-range-field/date-range-field.tsx
qroll: as discussed, the intended behaviour is:

- for range variant, disabled date between (exclusive) start/end date - ALLOWED
- for range variant, disabled date as start/end date - NOT ALLOWED
- the validation for week variant hasn't been finalised, so the future/past/min/max/disabled rules are ignored

weili-govtech: I see, in this case can we have an additional interface for the schema using the week variant? 
this new interface will exclude these validation rules and `disabledDates` and `allowDisabledSelection`

with this, devs will not be able to define the ignored rules / disable dates hence avoid confusion with its behaviour

src/stories/3-fields/date-range-field/date-range-field.stories.tsx
v1shva: Fixed

src/components/fields/date-range-field/date-range-field.tsx
v1shva: Fixed

v1shva: Fixed

src/stories/3-fields/date-range-field/date-range-field.stories.tsx
```diff
@@ -0,0 +1,217 @@
+import { DateTimeFormatter, LocalDate, ResolverStyle } from "@js-joda/core";
+import { Locale } from "@js-joda/locale_en-us";
+import { InputRangeProp } from "@lifesg/react-design-system/input-range-select/types";
+import { ArgsTable, Description, Heading, PRIMARY_STORY, Stories, Title } from "@storybook/addon-docs";
+import { Meta } from "@storybook/react/types-6-0";
```
weili-govtech, line 5:
```suggestion
import { Meta } from "@storybook/react";
```

weili-govtech (changes requested):

weili-govtech (approved):

PR #136, opened 2023-08-11
Storybook7

weili-govtech:
**Changes**
- Update to use Storybook 7
- Use storybook 7 typings
- Clean up frontend engine stories


src/stories/2-frontend-engine/frontend-engine.stories.tsx
```diff
@@ -386,10 +387,10 @@ export const OverrideSchema: Story<IFrontendEngineProps> = () => {
 	);
 };
 
-export const GetValues: Story<IFrontendEngineProps> = () => {
+export const GetValues: StoryFn<IFrontendEngineProps> = () => {
 	const ref = useRef<IFrontendEngineRef>();
 	const handleClick = () => {
-		console.log(ref.current.getValues());
+		action("getValues")(ref.current.getValues());
```
qroll, line 393: can update the button label as well as it's still referring to the console

.storybook/preview.js
```diff
@@ -15,7 +14,5 @@ export const parameters = {
 			order: ["Introduction", "Form", ["Frontend Engine", "Validation Schema", "Conditional Rendering"]],
 		},
 	},
+	decorators: [withA11y],
```
qroll, line 17: there's a warning `export 'withA11y' (imported as 'withA11y') was not found in '@storybook/addon-a11y'`

seems like registering the addon is sufficient

weili-govtech: I didn't get the warning actually but it appears there is no need to use the `withA11y` decorator indeed, I have removed the use of it

qroll (approved):

PR #139, opened 2023-08-15
Unregister conditionally rendered fields

qroll:
**Changes**

- Ensure default values of hidden fields at any level of nesting are excluded on change/submission by explicitly unregistering them
- `ConditionalRenderer` has to handle nested chip textarea
- The unregister logic for the textarea in `chips.tsx` has to stay because `ConditionalRenderer` can't handle the case where chips is shown, but textarea is not
- [delete] branch

**Changelog entry**

- Remove prefilled values of hidden fields from form state

### DIFF
--- a/src/components/elements/wrapper/conditional-renderer.tsx
+++ b/src/components/elements/wrapper/conditional-renderer.tsx
@@ -1,6 +1,7 @@
 import isEmpty from "lodash/isEmpty";
 import isObject from "lodash/isObject";
 import React, { useState } from "react";
+import { useFormContext } from "react-hook-form";
 import { useDeepCompareEffectNoCheck } from "use-deep-compare-effect";
 import { useFormValues, useValidationConfig } from "../../../utils/hooks";
 import { IFilterCheckboxSchema } from "../../custom/filter/filter-checkbox/types";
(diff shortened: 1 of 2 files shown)


src/components/elements/wrapper/conditional-renderer.tsx
```diff
@@ -80,8 +85,16 @@ export const ConditionalRenderer = ({ id, renderRules, children, schema }: IProp
...
 	 */
-	const listAllChildIds = (children: unknown) => {
+	const listAllChildIds = (schema: IProps["schema"]) => {
+		const children = schema["children"];
 		const childIdList: string[] = [];
+
+		// Handle special fields that render additional fields
+		if (schema.uiType === "chips") {
+			childIdList.push(id + "-textarea");
+		}
```
weili-govtech, line 95: why add chips textarea if the actual unregister is in chips component itself?

qroll: `Chips` cannot handle the case where it is nested and hidden, because it doesn't get mounted. So `ConditionalRenderer` has to handle this since it is aware of nested fields.

PR #148, opened 2023-08-31
Dirty fields

weili-govtech:
**Changes**
- Added the test cases for isDirty state for rest of the fields
- Split test cases for email and numeric fields into new files
- Misc refactoring of test cases

### DIFF
--- a/src/__tests__/components/fields/unit-number-field/unit-number.spec.tsx
+++ b/src/__tests__/components/fields/unit-number-field/unit-number.spec.tsx
@@ -1,9 +1,13 @@
 import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import cloneDeep from "lodash/cloneDeep";
+import merge from "lodash/merge";
 import { FrontendEngine } from "../../../../components";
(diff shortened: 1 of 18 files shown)


src/__tests__/components/fields/unit-number-field/unit-number.spec.tsx
```diff
@@ -161,4 +172,65 @@ describe(UI_TYPE, () => {
...
+			setFieldValue("12", "34");
+			fireEvent.click(screen.getByRole("button", { name: "Custom Button" }));
+
+			expect(formIsDirty).toBe(true);
+		});
+
+		it("should support default value without setting form state as dirty", () => {
+			render(
+				<FrontendEngineWithCustomButton
+					data={{ ...JSON_SCHEMA, defaultValues: { [COMPONENT_ID]: "hello" } }}
```
qroll, line 204: doesn't affect the result but the default value should be in the unit number format

weili-govtech: oops missed that out, I have fixed it

qroll (approved):

PR #159, opened 2023-09-29
[GETVALUES][WL] add payload argument to getValues()

weili-govtech:
**Changes**
- added `payload` argument to `getValues()`
- this follows the same behaviour as how react-hook-form retrieves values (https://react-hook-form.com/docs/useform/getvalues)


qroll (changes requested):
the typing of `getValues` in `src/components/frontend-engine/types.ts` will also need to be updated

weili-govtech:
oops added, I forgot to commit that change

qroll (approved):

PR #165, opened 2023-10-16
Refactor: Move imports to individual folders

weili-govtech:
**Changes**
- Move component-specific imports into their own folders (`fields`, `elements`, `custom-elements`)
- This reduces the size of the main types file
- New components will no longer require changes to the main file; changes will be limited to the corresponding folders
- There will be documentations on how to add components


src/components/custom/types.ts
```diff
@@ -40,18 +42,31 @@ export interface ICustomElementJsonSchema<T> {
...
-export interface IGenericCustomFieldProps<T = TFrontendEngineFieldSchema> {
+export interface IGenericCustomElementProps<T = TCustomComponentSchema> {
+	id: string;
+	schema: T;
+}
+
+/**
+ * common props for all custom fields
+ */
+export interface IGenericCustomFieldProps<T = IFilterCheckboxSchema>
```
qroll, line 67: can `T` be left as a required generic parameter? or is the default required?

weili-govtech: Ok I have removed it and did so for the other generic props as well because I realised these generic props are only used by the individual fields / elements / custom components.

qroll (approved):

PR #166, opened 2023-10-17
skip non-required validation for empty fields

weili-govtech:
**Changes**
- skip non-required validation for empty fields
- this fixes scenarios where non-required validation rules still run after emptying the fields


nobeltws (approved):
Oops, I saw the pipeline is failing some unit tests regarding assertions on expected error messages. Since now its returning undefined, there shouldn't be error thrown

nobeltws (approved):
thanks for the quick update to this validation checks~

PR #171, opened 2023-11-01
Divider and grid layout

weili-govtech:
**Changes**
- New `divider` element
- New `layoutType` for `section` to render grid layout

### DIFF
--- a/src/components/elements/divider/types.ts
+++ b/src/components/elements/divider/types.ts
@@ -1,0 +1,9 @@
+import { DividerProps } from "@lifesg/react-design-system/divider";
+import { TComponentOmitProps } from "../../frontend-engine";
+import { IBaseElementSchema } from "../types";
+
+export interface IDividerSchema
+	extends IBaseElementSchema<"divider">,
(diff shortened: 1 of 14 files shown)


src/components/elements/divider/types.ts
```diff
@@ -0,0 +1,9 @@
+import { DividerProps } from "@lifesg/react-design-system/divider";
+import { TComponentOmitProps } from "../../frontend-engine";
+import { IBaseElementSchema } from "../types";
+
+export interface IDividerSchema
+	extends IBaseElementSchema<"divider">,
+		TComponentOmitProps<DividerProps, "layoutType" | "desktopCols" | "tabletCols" | "mobileCols"> {
```
weili-govtech, line 7: note: I am excluding design system's grid implementation because frontend engine's wrapper already has a similar implementation, hence these props are not needed.

qroll (approved):

src/components/elements/divider/types.ts
qroll: just noting that this is a type-only omission, so if the consumer still specifies the props, it would still pass through to the underlying component. don't think it'll actually break anything though

PR #173, opened 2023-11-03
[CCUBE-556][DV] Location field enhancements

daiproxomhoa:
Location field requires a few enhancements:

Dispatch event when location modal is shown
Dispatch event when location modal is dismissed
Accept `disabled` key
Add unit tests and stories

### DIFF
--- a/src/components/fields/location-field/location-field.tsx
+++ b/src/components/fields/location-field/location-field.tsx
@@ -83,17 +83,16 @@ export const LocationField = (props: IGenericFieldProps<ILocationFieldSchema>) =
 		setValue(id, updatedValues, { shouldDirty });
 	};
 
-	const onFocus = (e?: React.FocusEvent<HTMLInputElement, Element>) => {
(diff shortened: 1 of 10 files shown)


src/components/fields/location-field/location-field.tsx
```diff
@@ -91,12 +108,11 @@ export const LocationField = (props: IGenericFieldProps<ILocationFieldSchema>) =
...
-					e.currentTarget.blur();
-				}}
+				onFocus={onFocus}
 				value={formValue?.address || ""}
 				errorMessage={error?.message}
+				disabled={disabled}
+				readOnly={readOnly}
 			/>
 			{!!formValue?.lat && !!formValue?.lng && (
 				<StyledStaticMap
```
weili-govtech, line 118: - StyledStaticMap should use `not-allowed` cursor if `disabled=true` 
- add `aria-disabled` to `src/components/shared/static-map/static-map.tsx` if component is disabled

weili-govtech (changes requested):

PR #175, opened 2023-11-05
Misc fixes

weili-govtech:
**Changes**
- turn off default international number validation
- update storybook packages to v7.5.2
- apply general error handling to provide more informative error when `data` or `sections` is not provided


qroll:
> turn off default international number validation

what's the rationale for this? thought this field would be similar to email where the input can be optional, but it's not empty and is an invalid format, an error should be raised

weili-govtech:
I am removing its default application because it is causing confusions as to what validation rule to apply, as it stands now, the international number validation rule has become pointless as it is applied by default. By removing this default behaviour, I am also allowing developers to apply their own validation if they don't want to use our one-size-fit-all international number validation.

qroll (approved):

PR #176, opened 2023-11-07
[CCUBE-547][DV] Update review component to support accordion variant

daiproxomhoa:
REQUIREMENTS

Ability to set half / full width values
Mask / unmask uinfin option (To confirm exact behaviour)
Optional button
Fire click event on clicking button
PROPOSED SCHEMA
{
    "referenceKey": "review",
    "variant": "accordion",
    "label": "Your personal information",
    "description": "Retrieved on 27 Jun 2023",
    "button": {
        "label": "Edit"
    },
    "collapsible":  true,
    "expanded": false,
    "items": [
        {
            "label": "Name (as in NRIC or passport)",
            "value": "Tom Tan Li Ho",
            "displayWidth": "half"
        },
        {
            "label": "NRIC or FIN",
            "value": "S••••534J",
            "displayWidth": "half"
        },
        {
            "label": "Date of birth",
            "value": "6 November 1992",
            "displayWidth": "half"
        },
        {
            "label": "Residential Address",
            "value": "Block 287, #05-11, Tampines street 22, Singapore 534788",
            "displayWidth": "full"
        },
        {
            "label": "Ethnicity",
            "value": "Chinese",
            "displayWidth": "half"
        }
    ]
}

### DIFF
--- a/src/components/custom/review/review.tsx
+++ b/src/components/custom/review/review.tsx
@@ -1,4 +1,5 @@
-import { BoxContainer, Button } from "@lifesg/react-design-system";
+import { BoxContainer } from "@lifesg/react-design-system/box-container";
+import { Button } from "@lifesg/react-design-system/button";
 import { UneditableSection } from "@lifesg/react-design-system/uneditable-section";
 import { useEffect } from "react";
 import styled from "styled-components";
@@ -6,12 +7,12 @@ import * as Yup from "yup";
 import { useFieldEvent, useValidationConfig } from "../../../utils/hooks";
 import { Wrapper } from "../../elements/wrapper";
 import { IGenericCustomElementProps } from "../types";
-import { IReviewSchema, IReviewSchemaDefault } from "./types";
+import { IReviewSchema, IReviewSchemaAccordion, IReviewSchemaDefault } from "./types";
 
 const UneditableSectionCustom = styled(UneditableSection)`
 	background-color: transparent;
-	padding-left: 2rem !important;
-	padding-right: 2rem !important;
+	padding-left: 2rem;
+	padding-right: 2rem;
 `;
 export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
 	// =============================================================================
@@ -48,15 +49,12 @@ export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
 	// =========================================================================
 	// RENDER FUNCTIONS
 	// =========================================================================
-	if (schema.variant === "accordion") {
-		const { items, button, label, expanded, collapsible, ...otherSchema } = schema;
+	const renderAccordion = (schema: IReviewSchemaAccordion) => {
+		const { items, button, label, ...otherSchema } = schema;
 		return (
 			<BoxContainer
-				id={id}
 				data-testid={id}
 				title={label}
-				expanded={expanded}
-				collapsible={collapsible}
 				callToActionComponent={
 					<Button.Default
 						styleType="light"
@@ -68,23 +66,27 @@ export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
 						{button?.label ?? "Edit"}
 					</Button.Default>
 				}
+				{...otherSchema}
 			>
-				<UneditableSectionCustom {...otherSchema} id={id} items={items} />
+				<UneditableSectionCustom id={id} items={items} />
 			</BoxContainer>
 		);
-	}
+	};
 
-	const { label, description, items, topSection, bottomSection, ...otherSchema } = schema;
+	const renderDefault = (schema: IReviewSchemaDefault) => {
+		const { label, description, items, topSection, bottomSection, ...otherSchema } = schema;
+		return (
+			<UneditableSection
+				{...otherSchema}
+				id={id}
+				title={label}
+				description={description}
+				items={items}
+				topSection={generateSection(topSection)}
+				bottomSection={generateSection(bottomSection)}
+			/>
+		);
+	};
 
-	return (
-		<UneditableSection
-			{...otherSchema}
-			id={id}
-			title={label}
-			description={description}
-			items={items}
-			topSection={generateSection(topSection)}
-			bottomSection={generateSection(bottomSection)}
-		/>
-	);
+	return schema.variant === "accordion" ? renderAccordion(schema) : renderDefault(schema);
 };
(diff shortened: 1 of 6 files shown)


src/components/custom/review/types.ts
```diff
@@ -2,13 +2,27 @@ import { UneditableSectionItemProps } from "@lifesg/react-design-system/uneditab
...
+	variant?: "default" | undefined;
+	label?: string | undefined;
+	description?: string | undefined;
+}
+export interface IReviewSchemaAccordion
+	extends IReviewBase,
+		Omit<BoxContainerProps, "children" | "callToActionComponent"> {
+	variant: "accodion";
+	button?: IButtonAccordion | undefined;
+}
```
weili-govtech, line 25: - we always use `label` for all schemas, so you can replace `title` with `label`
- `label` should be a required key
- due to ux requirements, we need to hide some options from the accordion variant because they is not meant to be used

```
export interface IReviewBase extends ICustomElementJsonSchema<"review"> {
	label: string;
	items: UneditableSectionItemProps[];
}
export interface IReviewSchemaDefault extends IReviewBase {
	variant?: "default" | undefined;
	description?: string | undefined;
	topSection?: Record<string, TReviewSectionChildren> | undefined;
	bottomSection?: Record<string, TReviewSectionChildren> | undefined;
}
export interface IReviewSchemaAccordion
	extends IReviewBase,
		Omit<BoxContainerProps, "children" | "title" | "callToActionComponent" | "subComponentTestIds"> {
	variant: "accordion";
	button?: IButtonAccordion | undefined;
}
```

src/components/custom/review/types.ts
```diff
@@ -2,13 +2,27 @@ import { UneditableSectionItemProps } from "@lifesg/react-design-system/uneditab
 import type { IAlertSchema, ITextSchema } from "../../elements";
 import type { IWrapperSchema } from "../../elements/wrapper";
 import type { ICustomElementJsonSchema } from "../types";
+import { BoxContainerProps } from "@lifesg/react-design-system";
```
weili-govtech, line 5:
```suggestion
import { BoxContainerProps } from "@lifesg/react-design-system/box-container";
```

weili-govtech, line 5: do organise the imports via vscode

src/components/custom/review/review.tsx
```diff
@@ -39,6 +45,40 @@ export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
...
+				/>
+			</BoxContainer>
+		);
+	}
+
+	const { label, description, items, topSection, bottomSection, ...otherSchema } = schema;
+
 	return (
 		<UneditableSection
 			{...otherSchema}
```
weili-govtech, line 84: can refactor to render each variant through functions
```
const renderAccordion = () => {
	// ...
};

const renderDefault = () => {
	// ...
};

return (schema.variant === "accordion") ? renderAccordion() : renderDefault();
```

src/components/custom/review/review.tsx
```diff
@@ -39,6 +45,40 @@ export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
...
+							dispatchFieldEvent("button-click", id);
+						}}
+					>
+						{button?.label ?? "Edit"}
+					</Button.Default>
+				}
+			>
+				<UneditableSectionCustom
+					{...otherSchema}
+					id={id}
```
weili-govtech, line 71: id is meant to be unique so it should not be applied to both BoxContainer and UneditableSection, can remove it.

src/components/custom/review/review.tsx
```diff
@@ -39,6 +45,40 @@ export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
...
+						onClick={() => {
+							dispatchFieldEvent("button-click", id);
+						}}
+					>
+						{button?.label ?? "Edit"}
+					</Button.Default>
+				}
+			>
+				<UneditableSectionCustom
+					{...otherSchema}
```
weili-govtech, line 70: schema is extended from `BoxContainerProps`, it has no overlap with `UneditableSection` props, hence there is no point spreading it here

src/components/custom/review/review.tsx
```diff
@@ -39,6 +45,40 @@ export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
...
 	// =========================================================================
+	if (schema.variant === "accodion") {
+		const { items, topSection, bottomSection, button, title, expanded, collapsible, ...otherSchema } = schema;
+		return (
+			<BoxContainer
+				id={id}
+				data-testid={id}
+				title={title}
+				expanded={expanded}
+				collapsible={collapsible}
```
weili-govtech, line 56: - schema supports `data-testid` so `id` should not be applied to it
- can spread `otherSchema` here instead of applying every prop manually

src/components/custom/review/review.tsx
```diff
@@ -1,26 +1,32 @@
...
 import { Wrapper } from "../../elements/wrapper";
 import { IGenericCustomElementProps } from "../types";
 import { IReviewSchema } from "./types";
-import { useValidationConfig } from "../../../utils/hooks";
-import { useEffect } from "react";
 
+const UneditableSectionCustom = styled(UneditableSection)`
+	background-color: transparent;
+	padding-left: 2rem !important;
+	padding-right: 2rem !important;
```
weili-govtech, line 14: no need to use `!important` flag

weili-govtech (changes requested):

weili-govtech:
note: updated this PR to merge to `review` branch instead because the grid column does not follow our design specifications, since you do not have access to the mocks, I will look into fixing it after you are done.

src/components/custom/review/types.ts
weili-govtech: do apply this fix

src/components/custom/review/review.tsx
```diff
@@ -39,6 +45,40 @@ export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
 	// =========================================================================
 	// RENDER FUNCTIONS
 	// =========================================================================
+	if (schema.variant === "accodion") {
+		const { items, topSection, bottomSection, button, title, expanded, collapsible, ...otherSchema } = schema;
+		return (
+			<BoxContainer
+				id={id}
+				data-testid={id}
```
weili-govtech, line 56: remove as schema has `data-testid`

src/components/custom/review/review.tsx
weili-govtech: remove applying id

PR #191, opened 2023-12-05
Misc fixes

weili-govtech:
**Changes**
- skip includes and excludes validation rules if value is empty
- enforce string label for button


nobeltws (approved):
thanks for the quick fix!

PR #193, opened 2023-12-06
[RBS-952] Nested multi-select component 

Mono143:
**Changes**
- Added a new component, nested multi-select.
- Added relevant tests and stories

-   [keep] branch

**Changelog entry**

-  Added new nested multi-select component

**Additional information**

-   You may refer to [RBS-952](https://jira.ship.gov.sg/browse/RBS-952)
-   Awaiting bugfixes in DS

### DIFF
--- a/src/components/fields/nested-multi-select/nested-multi-select.tsx
+++ b/src/components/fields/nested-multi-select/nested-multi-select.tsx
@@ -1,0 +1,89 @@
+import { Form } from "@lifesg/react-design-system/form";
+import { useEffect, useState } from "react";
+import { useFormContext } from "react-hook-form";
+import useDeepCompareEffect from "use-deep-compare-effect";
+import * as Yup from "yup";
+import { IGenericFieldProps } from "..";
+import { TestHelper } from "../../../utils";
+import { useValidationConfig } from "../../../utils/hooks";
+import { ERROR_MESSAGES } from "../../shared";
+import { ISelectOption } from "../select/types";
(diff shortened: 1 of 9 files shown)


src/components/fields/nested-multi-select/nested-multi-select.tsx
```diff
@@ -0,0 +1,154 @@
...
+	// =============================================================================
+	// CONST, STATE, REFS
+	// =============================================================================
+	const {
+		schema: { label, validation, options, ...otherSchema },
+		id,
+		value,
+		onChange,
+		error,
+		...otherProps
```
weili-govtech, line 23: we recently made some enhancements to the label, now all fields are injected with the `formattedLabel` prop, may I trouble you to add it as well, you can refer to: https://github.com/LifeSG/web-frontend-engine/blob/main/src/components/fields/text-field/text-field.tsx#L17

weili-govtech (changes requested):

qroll:
could I trouble you to rebase your changes on top of master instead? GitHub diff doesn't work well with merge commits

Mono143:
one thing to note: in the schema tests i noticed that adding a key to the overridden/updated options will cause the test to fail with a TypeError of `Cannot destructure property 'label' of 's' as it is undefined.` 
Can't seem to figure out why this happens. Aside from that everything is ready for review

qroll (changes requested):

qroll:
when testing overrides, I noticed that the DS component throws an error when the selected key path doesn't match any option. will be raising a PR to fix this

PR #194, opened 2023-12-06
Option label

weili-govtech:
**Changes**
- Allow html text for checkbox and radio button labels


nobeltws (approved):
thanks for the feature! 👍

PR #195, opened 2023-12-07
remove assertion of console.log in sanitize tests

weili-govtech:
**Changes**
- remove false positive assertion of js call in sanitize test cases

### DIFF
--- a/src/components/fields/checkbox-group/checkbox-group.tsx
+++ b/src/components/fields/checkbox-group/checkbox-group.tsx
@@ -11,6 +11,7 @@ import { useValidationConfig } from "../../../utils/hooks";
 import { ERROR_MESSAGES, Sanitize } from "../../shared";
 import { CheckboxContainer, Label, StyledCheckbox, ToggleWrapper } from "./checkbox-group.styles";
 import { ICheckboxGroupSchema } from "./types";
+import sanitize from "sanitize-html";
 
 export const CheckboxGroup = (props: IGenericFieldProps<ICheckboxGroupSchema>) => {
 	// =============================================================================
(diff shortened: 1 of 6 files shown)


qroll (commented):
can also assert that the script tag is stripped by checking `expect(div.innerHTML.includes("script")).toBe(false)`

src/components/fields/checkbox-group/checkbox-group.tsx
```diff
@@ -107,7 +108,7 @@ export const CheckboxGroup = (props: IGenericFieldProps<ICheckboxGroupSchema>) =
 							data-testid={TestHelper.generateId(id, "checkbox")}
 							id={checkboxId}
 							disabled={disabled ?? option.disabled}
-							name={option.label}
+							name={sanitize(option.label)}
```
qroll, line 111: name could take `option.value` instead since it represents the field value, rather than the display value

weili-govtech: decided to assign the unique id for the names instead, have applied the change to radio buttons as well

src/components/fields/checkbox-group/checkbox-group.tsx
```diff
@@ -11,6 +11,7 @@ import { useValidationConfig } from "../../../utils/hooks";
 import { ERROR_MESSAGES, Sanitize } from "../../shared";
 import { CheckboxContainer, Label, StyledCheckbox, ToggleWrapper } from "./checkbox-group.styles";
 import { ICheckboxGroupSchema } from "./types";
+import sanitize from "sanitize-html";
```
qroll, line 14: can remove this unused import

qroll (approved):

PR #197, opened 2023-12-12
default filter expanded to true

weili-govtech:
**Changes**
- Update filter checkbox and filter item `expanded` prop to be true by default

**Changelog entry**
-   `filter-item` and `filter-checkbox` are now expanded by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The intention of frontend engine is to take out the heavy lifting of form develo
 
 Developers are expected to have the following packages installed:
 
--   @lifesg/react-design-system 2.3.0-canary.2
+-   @lifesg/react-design-system 2.3.0-canary.3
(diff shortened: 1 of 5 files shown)


qroll (dismissed):

qroll:
code changes look fine but I don't have context on why the default is changed. for the PS use case, those filters are closed by default 🤔

qroll:
thanks @xyukianesa for sharing on the `expanded` bug in mobile view

this is an issue on DS side, I did not check if the item is mobile before applying the prop in [this change](https://github.com/LifeSG/react-design-system/pull/327). putting up a PR to fix this.

weili-govtech:
thanks for the DS fix, I have replaced the changes with a DS update and changed the expanded description in storybook to mention it is applicable to desktops only

qroll (approved):

PR #201, opened 2023-12-27
Slider

qroll:
**Changes**

- Add slider component from DS
- Support validation through existing `min` and `max` rules, and a custom rule called `increment` that restricts the interval of values
  - if increment = 2, allowed values are 0, 2, 4, etc.
  - if min = 1.5, increment = 0.5, allowed values are 1.5, 2, 2.5, etc.
  - these rules will map to the underlying slider props
-   [delete] branch

**Changelog entry**

-  New slider field

### DIFF
--- a/src/components/fields/slider/slider.tsx
+++ b/src/components/fields/slider/slider.tsx
@@ -1,0 +1,87 @@
+import { Form } from "@lifesg/react-design-system/form";
+import { useEffect, useState } from "react";
+import * as Yup from "yup";
+import { IGenericFieldProps } from "..";
+import { TestHelper } from "../../../utils";
+import { useValidationConfig } from "../../../utils/hooks";
+import { ISliderSchema } from "./types";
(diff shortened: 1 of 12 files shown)


src/components/fields/slider/slider.tsx
```diff
@@ -0,0 +1,87 @@
...
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [validation]);
+
+	// =============================================================================
+	// EVENT HANDLERS
+	// =============================================================================
+	const handleChange = (value: number): void => {
+		onChange({ target: { value } });
+	};
```
weili-govtech, line 66: the change event should be fired only after user confirmed the value aka on drag end instead of firing as the user drags the knob, this is based on the HTML range input's behaviour

weili-govtech (changes requested):

weili-govtech (approved):

PR #207, opened 2024-01-12
Histogram slider

qroll:
**Changes**
- Port HistogramSlider, bumped DS to `2.3.0-canary.4` accordingly
- Value in the shape of `{ from: number; to: number; }`
- Support implicit validation through the `bins` and `interval` prop
-   [delete] branch

**Changelog entry**

-   New histogram slider field

### DIFF
--- a/src/__tests__/components/fields/histogram-slider/histogram-slider.spec.tsx
+++ b/src/__tests__/components/fields/histogram-slider/histogram-slider.spec.tsx
@@ -1,0 +1,247 @@
+import { Form } from "@lifesg/react-design-system/form";
+import { FormHistogramSliderProps } from "@lifesg/react-design-system/form/types";
+import { act, fireEvent, render, screen, waitFor } from "@testing-library/react";
+import cloneDeep from "lodash/cloneDeep";
+import merge from "lodash/merge";
+import { FrontendEngine } from "../../../../components";
+import { IHistogramSliderSchema } from "../../../../components/fields";
+import { IFrontendEngineData, IFrontendEngineRef } from "../../../../components/frontend-engine";
+import { ERROR_MESSAGES } from "../../../../components/shared";
+import {
+	ERROR_MESSAGE,
+	FRONTEND_ENGINE_ID,
+	FrontendEngineWithCustomButton,
+	TOverrideField,
+	TOverrideSchema,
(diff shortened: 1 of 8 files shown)


src/components/fields/histogram-slider/types.ts
```diff
@@ -0,0 +1,21 @@
...
+	from: number | undefined;
+	to: number | undefined;
+}
+
+export interface IHistogramSliderSchema<V = undefined>
+	extends IBaseFieldSchema<"histogram-slider", V>,
+		TComponentOmitProps<
+			Pick<
+				HistogramSliderProps,
+				"className" | "id" | "data-testid" | "disabled" | "readOnly" | "bins" | "interval"
```
weili-govtech, line 17: `id` is not needed

src/stories/3-fields/histogram-slider/histogram-slider.stories.tsx
```diff
@@ -0,0 +1,211 @@
...
+	label: "Price of fruits",
+	bins: [
+		{ minValue: 0, count: 0 },
+		{ minValue: 10, count: 2 },
+		{ minValue: 20, count: 3 },
+		{ minValue: 90, count: 8 },
+	],
+	interval: 10,
+	validation: [{ required: true }],
+};
```
weili-govtech, line 148: after trying out this behaviour, I realise we need to clarify with UX. there may be some confusions here because by default the field is showing a range selected but it is still asking the user to make a selection

src/__tests__/components/fields/histogram-slider/histogram-slider.spec.tsx
```diff
@@ -0,0 +1,239 @@
...
+const UI_TYPE = "histogram-slider";
+
+const JSON_SCHEMA: IFrontendEngineData = {
+	id: FRONTEND_ENGINE_ID,
+	sections: {
+		section: {
+			uiType: "section",
+			children: {
+				[COMPONENT_ID]: {
+					label: "Select",
```
weili-govtech, line 32: can change to `Histogram Slider`

weili-govtech (changes requested):

weili-govtech (approved):

PR #212, opened 2024-01-22
File upload

weili-govtech:
**Changes**
- New file upload field based on [DS FileUpload component](https://designsystem.life.gov.sg/react/index.html?path=/docs/modules-fileupload--docs)
- Structurally similar to existing `image-upload` field because `file-upload` field will be enhanced to cover `image-upload` field in the future
- Supports `base64` and `multipart` uploads

### DIFF
--- a/src/components/fields/file-upload/file-upload-manager.ts
+++ b/src/components/fields/file-upload/file-upload-manager.ts
@@ -1,0 +1,321 @@
+import { useContext, useEffect, useRef } from "react";
+import { useFormContext } from "react-hook-form";
+import { AxiosApiClient, FileHelper, ImageHelper } from "../../../utils";
+import { usePrevious } from "../../../utils/hooks";
+import { ERROR_MESSAGES } from "../../shared";
+import { FileUploadContext } from "./file-upload-context";
+import { EFileStatus, IFile, IFileUploadSchema, IFileUploadValidationRule, IFileUploadValue } from "./types";
+
+interface IProps {
+	fileTypeRule: IFileUploadValidationRule;
+	id: string;
+	maxFileSizeRule: IFileUploadValidationRule;
+	upload: IFileUploadSchema["uploadOnAddingFile"];
+	value: IFileUploadValue[];
+}
+
+export const FileUploadManager = (props: IProps) => {
+	// =============================================================================
+	// CONST, STATE, REFS
+	// =============================================================================
+	const { fileTypeRule, id, maxFileSizeRule, upload, value } = props;
+	const { files, setFiles } = useContext(FileUploadContext);
+	const previousValue = usePrevious(value);
+	const { setValue } = useFormContext();
+	const sessionId = useRef<string>();
+
+	// =============================================================================
+	// EFFECTS
+	// =============================================================================
+	useEffect(() => {
+		sessionId.current = generateRandomId();
+	}, []);
+
+	useEffect(
+		() => {
+			// handle file status
+			files.forEach(async (file, index) => {
+				try {
+					switch (file.status) {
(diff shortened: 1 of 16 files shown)


src/components/fields/file-upload/file-upload.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+	// =============================================================================
+	// EVENT HANDLERS
+	// =============================================================================
+	const handleChange = (rawFiles: File[]) => {
+		handleNewFiles(
+			rawFiles.map((rawFile) => ({
+				addedFrom: "input" as const,
+				rawFile,
+				slot: FileUploadHelper.findAvailableSlot(files),
```
qroll, line 148: the same slot is assigned when multiple files are uploaded at the same time, as the files array is not the latest

anyway looking through the manager and component, I don't see the use of slot now as each file item now has an id?

src/components/fields/file-upload/file-upload.tsx
```diff
@@ -0,0 +1,241 @@
...
+		id,
+		error,
+		isDirty,
+		value,
+		schema: { description, label, uploadOnAddingFile, validation, warning, ...otherSchema },
+	} = props;
+	const { files, setFiles } = useContext(FileUploadContext);
+	const fileTypeRuleRef = useRef<IFileUploadValidationRule>({});
+	const maxFilesRuleRef = useRef<IYupValidationRule>({});
+	const maxFileSizeRuleRef = useRef<IFileUploadValidationRule>({});
```
qroll, line 29: just curious, what's the use of putting these rules in a ref? setting ref doesn't trigger rerenders, although for this case it works because it's caused from a change in `validation` prop, which sets state and thus guarantees to trigger a rerender anyway

src/components/fields/file-upload/file-upload-manager.ts
```diff
@@ -0,0 +1,321 @@
...
+				fileItem: {
+					errorMessage,
+					id: generateRandomId(),
+					name: FileHelper.deduplicateFileName(
+						files.map(({ fileItem }) => fileItem?.name),
+						index,
+						fileToParse.rawFile.name
+					),
+					size: fileToParse.rawFile.size,
+					type: fileType.mime,
```
qroll, line 244: this line throws an error when a plain text file is uploaded. `fileType` is undefined somehow

qroll (changes requested):

src/components/fields/file-upload/file-upload.tsx
weili-govtech: ok have fixed the slot assignment. 

the slots are from the `image-upload` component and it is used by the OS track, I am porting this feature over first

src/components/fields/file-upload/file-upload.tsx
weili-govtech: I had to change to refs because at the time of initialisation, right after the rules are defined in the `useEffect`, they are needed as part of handling default values. because the states are not getting updated fast enough, I switched to refs to get around this issue.

src/components/fields/file-upload/file-upload-manager.ts
weili-govtech: it appears not to be possible to determine the file type of text-based files through magic numbers, the explanation is such files don't come with any signatures: https://github.com/sindresorhus/file-type/issues/264

in view of that, I have added a fall back to use whatever is defined in the file if the detection fails.

qroll:
some of the unit tests for dirty state are failing

weili-govtech:
Reverted my last commit because it was erroneously dirtying the form on reset

qroll (approved):

PR #225, opened 2024-02-26
[BOOKINGSG-5240][WS|GC] text and email fields to include uppercase custom option

gohguancheng:
**Changelog entry**

-   Adding custom option to allow text and email inputs to be uppercase

### DIFF
--- a/src/components/fields/text-field/text-field.tsx
+++ b/src/components/fields/text-field/text-field.tsx
@@ -12,15 +12,8 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
 	// ================================================
 	// CONST, STATE, REFS
 	// ================================================
-	const {
-		error,
-		formattedLabel,
-		id,
-		onChange,
-		value,
-		schema: { customOptions, inputMode, label: _label, uiType, validation, ...otherSchema },
-		...otherProps
-	} = props;
+	const { error, formattedLabel, id, onChange, value, schema, ...otherProps } = props;
+	const { customOptions, inputMode, label: _label, uiType, validation, ...otherSchema } = schema;
 
 	const [stateValue, setStateValue] = useState<string | number>(value || "");
 	const [derivedAttributes, setDerivedAttributes] = useState<FormInputProps>({});
@@ -95,7 +88,7 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
 	}, [value]);
 
 	useEffect(() => {
-		if (props.schema.uiType === "text-field" && ref.current.selectionEnd !== caret.current) {
+		if (uiType === "text-field" && ref.current.selectionEnd !== caret.current) {
 			// keep caret in place after uppercase, not available for 'email' & 'number' HTML input types
 			ref.current.setSelectionRange(caret.current, caret.current);
 		}
@@ -103,15 +96,11 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
 	}, [stateValue]);
 
 	useEffect(() => {
-		if (
-			props.schema.uiType === "text-field" &&
-			props.schema.customOptions?.textTransform === "uppercase" &&
-			stateValue
-		) {
-			setStateValue((previous) => (previous ? ("" + previous).toUpperCase() : ""));
+		if (schema.uiType === "text-field" && schema.customOptions?.textTransform === "uppercase" && stateValue) {
+			setStateValue((previous) => previous?.toString().toUpperCase());
 		}
 		// eslint-disable-next-line react-hooks/exhaustive-deps
-	}, [props.schema.uiType === "text-field" && props.schema.customOptions?.textTransform]);
+	}, [schema.uiType === "text-field" && schema.customOptions?.textTransform]);
 
 	// =============================================================================
 	// EVENT HANDLERS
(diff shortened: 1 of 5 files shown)


src/components/fields/text-field/text-field.tsx
```diff
@@ -98,6 +98,11 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
 		if (uiType === "numeric-field" && !event.target.value) {
 			onChange({ target: { value: undefined } });
 		} else {
+			const inputValue = event.target.value;
+			event.target.value =
+				customOptions?.uppercase && inputValue !== inputValue.toUpperCase()
```
weili-govtech, line 103: why the need to compare against an uppercased version instead of outrightly converting to uppercase?

since numeric field doesn't support uppercase, there is a need to ensure its custom options do not accept that

src/components/fields/text-field/types.ts
```diff
@@ -5,6 +5,7 @@ import { IBaseFieldSchema } from "../types";
 type TCustomOptions = {
 	preventCopyAndPaste?: boolean | undefined;
 	preventDragAndDrop?: boolean | undefined;
+	uppercase?: boolean | undefined;
```
weili-govtech, line 8: may I suggest to offer `textTransform` that can accept `uppercase` or `undefined`? this is just in case there is a need to offer other transformation in the future

src/components/fields/text-field/text-field.tsx
weili-govtech, line 1: there is also the need to convert to uppercase if uppercase got updated at runtime, e.g. via the `overrides` key, you can refer to the `overrides` story for more info

weili-govtech (changes requested):

src/components/fields/text-field/text-field.tsx
wsee911: we ensure only text-fields run the uppercase only through a check
and also typing the custom options

src/components/fields/text-field/types.ts
wsee911: updated

src/components/fields/text-field/text-field.tsx
```diff
@@ -91,12 +95,30 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [value]);
 
+	useLayoutEffect(() => {
```
weili-govtech, line 98: avoid `useLayoutEffect` because it will throw a warning on nextjs

src/components/fields/text-field/types.ts
```diff
@@ -2,15 +2,24 @@ import { FormInputProps } from "@lifesg/react-design-system/form/types";
...
 import { IBaseFieldSchema } from "../types";
 
+export enum ETextTransform {
+	UPPERCASE = "uppercase",
+}
+
 type TCustomOptions = {
 	preventCopyAndPaste?: boolean | undefined;
 	preventDragAndDrop?: boolean | undefined;
+	textTransform?: ETextTransform | undefined;
```
weili-govtech, line 12: instead of enum, can we have a union of the possible values? the reason is the devs will not have access to this enum when defining the schema

```suggestion
	textTransform?: "uppercase" | undefined;
```

src/components/fields/text-field/text-field.tsx
weili-govtech: notice if I have some values in the field and they do not get uppercased if the texttransform is changed to uppercase, this needs another useEffect

in addition, there is a need to track the original value keyed in by user, in case the textTransform reverts from `uppercase` to undefined

```
useEffect(() => {
	// apply text transform accordingly
}, [customOptions?.textTransform]);
```

weili-govtech (changes requested):

src/components/fields/text-field/types.ts
wsee911: ooo so thats why enum is rarely used from certain things ok

src/components/fields/text-field/types.ts
gohguancheng: updated and resolving

src/components/fields/text-field/text-field.tsx
gohguancheng: removed uppercase check. 

was needed on bsg side because FEE form onchange triggers on updating control values, resulting in infinite loop.

src/components/fields/text-field/text-field.tsx
gohguancheng: For future reference, we decided to keep to 1-way transformation at runtime to prevent overcomplicating the component for now.

Documentation updated to explain that input values will not revert when `textTransform` is unset (to `undefined`)

src/components/fields/text-field/text-field.tsx
gohguancheng: removed, using useEffect instead

src/components/fields/text-field/text-field.tsx
```diff
@@ -91,12 +94,42 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
...
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [stateValue]);
+
+	useEffect(() => {
+		if (
+			props.schema.uiType === "text-field" &&
+			props.schema.customOptions?.textTransform === "uppercase" &&
+			stateValue
+		) {
+			setStateValue((previous) => (previous ? ("" + previous).toUpperCase() : ""));
```
weili-govtech, line 111:
```suggestion
			setStateValue((previous) => previous?.toString().toUpperCase());
```

src/components/fields/text-field/text-field.tsx
```diff
@@ -91,12 +94,42 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
...
 	// =============================================================================
 	// EVENT HANDLERS
 	// =============================================================================
 	const handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
 		if (uiType === "numeric-field" && !event.target.value) {
 			onChange({ target: { value: undefined } });
+		} else if (uiType === "text-field") {
+			const inputted = event.target.value;
+			const doUpperCase =
+				props.schema.uiType === "text-field" && props.schema.customOptions?.textTransform === "uppercase";
```
weili-govtech, line 125: - type inference will also work if you use the full path for the uiType in the if condition
- can refactor a bit to refer via the schema instead of the props

```suggestion
		} else if (schema.uiType === "text-field") {
			const inputted = event.target.value;
			const doUpperCase = schema.customOptions?.textTransform === "uppercase";
```

weili-govtech (changes requested):
logic looks good, only left a couple minor refactoring to do

src/components/fields/text-field/text-field.tsx
gohguancheng: noted and applied accordingly.

src/components/fields/text-field/text-field.tsx
gohguancheng: interesting. thanks for sharing this. have used the above to simplify the code.

weili-govtech (approved):

PR #232, opened 2024-03-04
Support nested fields in checkbox and radio toggles

qroll:
**Changes**

- similar handling was implemented in validation-schema-generator: https://github.com/LifeSG/validation-schema-generator/pull/45
-   [delete] branch

**Changelog entry**

-   Support nested fields in checkbox and radio toggles

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-984)

### DIFF
--- a/src/components/fields/checkbox-group/types.ts
+++ b/src/components/fields/checkbox-group/types.ts
@@ -39,3 +39,6 @@ interface ICheckboxGroupToggleSchema<V = undefined>
 }
 
 export type TCheckboxGroupSchema<V = undefined> = ICheckboxGroupDefaultSchema<V> | ICheckboxGroupToggleSchema<V>;
+
+/** @deprecated will be removed in a future release. Use `TCheckboxGroupSchema` instead */
+export type ICheckboxGroupSchema<V = undefined> = ICheckboxGroupDefaultSchema<V> | ICheckboxGroupToggleSchema<V>;
--- a/src/components/fields/checkbox-group/types.ts
+++ b/src/components/fields/checkbox-group/types.ts
@@ -38,4 +38,4 @@ interface ICheckboxGroupToggleSchema<V = undefined>
 	};
 }
 
-export type ICheckboxGroupSchema<V = undefined> = ICheckboxGroupDefaultSchema<V> | ICheckboxGroupToggleSchema<V>;
+export type TCheckboxGroupSchema<V = undefined> = ICheckboxGroupDefaultSchema<V> | ICheckboxGroupToggleSchema<V>;
--- a/src/components/fields/checkbox-group/types.ts
+++ b/src/components/fields/checkbox-group/types.ts
@@ -1,5 +1,5 @@
 import { CheckboxProps } from "@lifesg/react-design-system/checkbox";
-import { TComponentOmitProps } from "../../frontend-engine";
+import { TComponentOmitProps, TFrontendEngineFieldSchema } from "../../frontend-engine";
 import { IBaseFieldSchema } from "../types";
 
 interface IOption {
@@ -10,6 +10,7 @@ interface IOption {
 
 interface IToggleOption extends IOption {
 	none?: boolean;
+	children?: Record<string, TFrontendEngineFieldSchema>;
 }
 
 export type TCheckboxToggleLayoutType = "horizontal" | "vertical";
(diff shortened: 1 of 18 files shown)


src/components/fields/checkbox-group/types.ts
```diff
@@ -10,6 +10,7 @@ interface IOption {
 
 interface IToggleOption extends IOption {
 	none?: boolean;
+	children?: Record<string, TFrontendEngineFieldSchema>;
```
weili-govtech, line 13: - add `undefined` typing
- I just noticed the toggle typing is not applied correctly, can we differentiate the option typing for toggle styleType? so it will reject children if styleType is default

```suggestion
	none?: boolean | undefined;
	children?: Record<string, TFrontendEngineFieldSchema> | undefined;
```

src/components/fields/radio-button/types.ts
```diff
@@ -1,12 +1,13 @@
...
-import { TComponentOmitProps } from "../../frontend-engine";
+import { TComponentOmitProps, TFrontendEngineFieldSchema } from "../../frontend-engine";
 import { IBaseFieldSchema } from "../types";
 
 interface IOption {
 	label: string;
 	value: string;
 	disabled?: boolean | undefined;
 	imgSrc?: string | undefined;
+	children?: Record<string, TFrontendEngineFieldSchema>;
```
weili-govtech, line 10: similar changes as checkbox
- add undefined type
- need to differentiate option type between toggle styleType and others

src/stories/3-fields/checkbox-group/checkbox-toggle-group.stories.tsx
weili-govtech, line 1: include children in the options prop description in props table

src/stories/3-fields/radio-button/radio-toggle-button.stories.tsx
weili-govtech, line 1: include children in the options prop description in props table

weili-govtech (changes requested):

src/components/fields/checkbox-group/types.ts
qroll: > I just noticed the toggle typing is not applied correctly, can we differentiate the option typing for toggle styleType? so it will reject children if styleType is default

after some experimentation, found that the `styleType` is nested, so Typescript is not able to perform type narrowing with discriminated union

I've also tried:
- set options to `IToggleOption[] | IOption[]` - not that helpful
- define the additional options like `customOptions.additionalOptions: IToggleOption[]` but that would decouple it from the main options

weili-govtech: I experimented a bit and found a way to make it work
- the schema for each styleType will need to be defined separately
- the final schema will be a type and not an interface anymore
```
interface ICheckboxGroupToggleSchema<V = undefined>
	extends IBaseFieldSchema<"checkbox", V>,
		TComponentOmitProps<CheckboxProps> {
	options: IToggleOption[];
	customOptions?:
		| {
				styleType: "toggle";
				indicator?: boolean | undefined;
				border?: boolean | undefined;
				layoutType?: TCheckboxToggleLayoutType | undefined;
		  }
		| undefined;
}

interface ICheckboxGroupDefaultSchema<V = undefined>
	extends IBaseFieldSchema<"checkbox", V>,
		TComponentOmitProps<CheckboxProps> {
	options: IOption[];
	customOptions?:
		| {
				styleType: "default";
		  }
		| undefined;
}

export type TCheckboxGroupSchema<V = undefined> = ICheckboxGroupToggleSchema<V> | ICheckboxGroupDefaultSchema<V>;
```

qroll: oh yes, this works. thanks for looking into this too

will the renaming of the interface be a breaking change?

weili-govtech: yeah I found that some projects are referencing the schema typing so it is going to break some projects

in this case, let's retain the old names yet change them to types. we will mark them as deprecated, I will announce their deprecation in the upcoming release and remove it in a future release

for the new typings, we will use them in place of the existing ones in `TFieldSchema`

src/components/fields/types.ts
```diff
@@ -66,7 +66,7 @@ export enum EFieldType {
  */
 export type TFieldSchema<V = undefined> =
 	| IButtonSchema
-	| ICheckboxGroupSchema<V>
+	| TCheckboxGroupSchema<V>
```
weili-govtech, line 69: can sort by alphabetical order?

weili-govtech (approved):

PR #235, opened 2024-03-07
create hidden field

weili-govtech:
**Changes**
- new `hidden-field` that renders hidden input
- comes with validation and conditional rendering but no label
- delete branch


src/components/fields/index.ts
```diff
@@ -4,8 +4,9 @@ export * from "./chips";
 export * from "./contact-field";
 export * from "./date-field";
 export * from "./date-range-field";
-export * from "./file-upload";
+export * from "./hidden-field";
 export * from "./histogram-slider";
+export * from "./file-upload";
```
qroll, line 9: was moving `file-upload` intended? as it was already in the correct alphabetical order

src/components/fields/hidden-field/hidden-field.tsx
```diff
@@ -0,0 +1,30 @@
...
+		schema: { showIf: _showIf, uiType: _uiType, validation, ...otherSchema },
+		...otherProps
+	} = props;
+	const { setFieldValidationConfig } = useValidationConfig();
+
+	// =============================================================================
+	// EFFECTS
+	// =============================================================================
+	useEffect(() => {
+		setFieldValidationConfig(id, Yup.string(), validation);
```
qroll, line 22: something to consider as a future enhancement is the ability to handle other types. a pain point I've encountered is having to cast boolean to string.

src/components/fields/index.ts
weili-govtech: whoops not intended, I have fixed it

src/components/fields/hidden-field/hidden-field.tsx
weili-govtech: oh yeah thanks for the suggestion, I will introduce a new validation rule for hidden field to validate as other data types

qroll (approved):

PR #239, opened 2024-03-12
Mol 14970

ryhuz:
**Changes**
- Added flag `disableLocationSelectionOnStart` to prevent default selection based on geolocation, and prevent map click
- Added dispatch event `current-location-has-set` when geolocation is fetched
- Added event listener `set-selectable-pins` for setting selectable pins
- Populate results list from selectable pins
- Allow selection of selectable pins

**[delete] branch

**Changelog entry**

-   Added flag `disableLocationSelectionOnStart`
-   Added events for setting selectable pins

**Additional information**

-   You may refer to this [MOL-14970](https://jira.ship.gov.sg/browse/MOL-14970)


src/components/fields/location-field/location-modal/location-picker/location-picker.tsx
```diff
@@ -131,23 +143,61 @@ export const LocationPicker = ({
...
 	};
 
-	const zoomWithMarkers = (target: ILocationCoord) => {
+	const zoomWithMarkers = (
+		targets: ILocationCoord[],
+		shouldSetMarkers = true,
+		_zoomCenter?: ILocationCoord,
+		selectableMarkers = false,
+		enlargeSelectedMarker = false
+	) => {
```
Huizz, line 152: would there be a case where shouldSetMarkers is false, and selectableMarkers and enlargeSelectedMarker are also false?

Means dont show current location marker
yet there are no selectable pins.

what can the user do in this state?

ryhuz: `enlargeSelectedMarker` shouldn't make a difference

`shouldSetMarkers` states whether there will be pins on the map

`selectableMarkers` state whether user can click on the pin

so if `shouldSetMarkers` is false, `selectableMarkers` won't make a difference since there is nothing to click anyway.

a separate setting in location-picker is whether or not user can click on map to select, so there is that option to select location

src/components/fields/location-field/location-modal/location-picker/location-picker.tsx
```diff
@@ -112,12 +115,21 @@ export const LocationPicker = ({
...
-		if (selectedLocationCoord?.lat && selectedLocationCoord?.lng) {
-			zoomWithMarkers(selectedLocationCoord);
-		} else {
+		if (!selectedLocationCoord?.lat || !selectedLocationCoord?.lng) {
 			resetView();
+			return;
+		}
+
+		if (selectablePins.length) {
+			const pins = selectablePins;
```
weili-govtech, line 124:
```suggestion
			const pins = [...selectablePins];
```

src/components/fields/location-field/location-modal/location-search/location-search.tsx
```diff
@@ -625,20 +649,21 @@ export const LocationSearch = ({
 						onFocus={handleInputFocus}
 						onChange={handleInputChange}
 						placeholder={addressFieldPlaceholder}
-						readOnly={gettingCurrentLocation}
+						readOnly={gettingCurrentLocation || restrictLocationSelection}
 						value={!gettingCurrentLocation ? queryString : gettingCurrentLocationFetchMessage}
 						ref={inputRef}
 						disabled={disableTextSearch}
```
weili-govtech, line 655: now that we have the need to set the input as readOnly and another prop to disable the text search, may I double confirm both props are needed?

I was under the impression `disableTextSearch` is meant for this but it appears that their behaviours are slightly different, in view of this, what is the use of `disableTextSearch`?

src/components/fields/location-field/types.ts
```diff
@@ -17,7 +18,8 @@ export interface ILocationFieldSchema<V = undefined>
 			| "disableTextSearch"
 		>,
 		Pick<ILocationInputProps, "locationInputPlaceholder" | "disabled" | "readOnly">,
-		Pick<IStaticMapProps, "staticMapPinColor"> {
+		Pick<IStaticMapProps, "staticMapPinColor">,
+		Pick<ILocationModalProps, "disableLocationSelectionOnStart"> {
```
weili-govtech, line 22: as `disableLocationSelectionOnStart` is applying a number of behaviours, may I suggest to introduce `selectionMode` instead? it will accept `"default" | "pins" | "undefined"`  with `pins` being the new behaviour you are introducing

weili-govtech (changes requested):

src/components/fields/location-field/location-modal/location-picker/location-picker.tsx
ryhuz: will update

src/components/fields/location-field/types.ts
ryhuz: updated

src/components/fields/location-field/location-modal/location-search/location-search.tsx
ryhuz: updated

weili-govtech (changes requested):

src/stories/3-fields/location-field/location-field.stories.tsx
```diff
@@ -73,17 +73,31 @@ const meta: Meta = {
...
+			description: "Specifies if the search input field is disabled or set to read only. Enabled if undefined",
+			table: {
+				type: {
+					summary: "string",
+				},
+				defaultValue: { summary: undefined },
+			},
+			options: ["disabled", "readonly", undefined],
+			control: {
+				type: "text",
```
ryhuz, line 100:
```suggestion
				type: "select",
```

src/stories/3-fields/location-field/location-field.stories.tsx
```diff
@@ -73,17 +73,31 @@ const meta: Meta = {
...
+					summary: "string",
 				},
-				defaultValue: { summary: false },
+				defaultValue: { summary: "default" },
 			},
-			options: [true, false],
+			options: ["default", "pins-only", undefined],
 			control: {
-				type: "boolean",
+				type: "text",
```
ryhuz, line 87:
```suggestion
				type: "select",
```

src/stories/3-fields/location-field/location-field.stories.tsx
ryhuz, line 100:
```suggestion
				type: "select",
			},
		},
		disableSearch: {
			description: "Specifies if the search input field is disabled or set to read only. Enabled if undefined",
			table: {
				type: {
					summary: "string",
				},
				defaultValue: { summary: undefined },
			},
			options: ["disabled", "readonly", undefined],
			control: {
				type: "select",
```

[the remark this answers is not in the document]
ryhuz: updated

weili-govtech (approved):

PR #252, opened 2024-04-01
Add DS TextList

qroll:
**Changes**

-   [delete] branch
- Support ordered and unordered lists
- Support list item of type `Record<string, TFrontendEngineFieldSchema>` for more complex formatting e.g. nested lists, other elements

DS component: https://designsystem.life.gov.sg/react/index.html?path=/docs/general-textlist--docs


src/components/elements/list/list.tsx
```diff
@@ -0,0 +1,44 @@
...
+		id,
+		schema: { children, uiType, ...otherSchema },
+	} = props;
+
+	const Element = uiType === "ordered-list" ? DSTextList.Ol : DSTextList.Ul;
+
+	// =============================================================================
+	// RENDER FUNCTIONS
+	// =============================================================================
+	const renderText = (): JSX.Element[] | JSX.Element => {
```
weili-govtech, line 22:
```suggestion
	const renderChildren = (): JSX.Element[] | JSX.Element => {
```

src/components/elements/types.ts
```diff
@@ -49,11 +53,15 @@ export enum EElementType {
 export type TElementSchema =
 	| IAlertSchema
 	| IDividerSchema
-	| ITextSchema
-	| IWrapperSchema
 	| IGridSchema
+	| IListItemSchema
```
weili-govtech, line 57: remove as `list-item` can only be added as a direct descendant of `ordered-list` and `unordered-list`

weili-govtech (changes requested):

src/components/elements/types.ts
qroll: hmm but this needs to be passed down to `Wrapper` (for cond rend handling), which requires the ui type to be known

weili-govtech: in such cases, we pass them as `children` so we will need to add `IListItemSchema` typing to the children prop
ref: https://github.com/LifeSG/web-frontend-engine/blob/main/src/components/elements/wrapper/types.ts#L20

weili-govtech (approved):

PR #260, opened 2024-04-16
[MOL-15419][NL] Add reset-location event

namlh023:
**Changes**

- Fire reset-location event when user click current location button
- Delete branch

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-15419)

### DIFF
--- a/src/__tests__/components/fields/location-field/location-field.spec.tsx
+++ b/src/__tests__/components/fields/location-field/location-field.spec.tsx
@@ -64,6 +64,8 @@ enum ELocationInputEvents {
 	"BEFORE_HIDE_PERMISSION_MODAL" = "before-hide-permission-modal",
 	"HIDE_PERMISSION_MODAL" = "hide-permission-modal",
 	"DISMISS_LOCATION_MODAL" = "dismiss-location-modal",
+	"RESET_LOCATION" = "reset-location",
+	"CONFIRM_RESET_LOCATION" = "confirm-reset-location",
 }
 interface ICustomFrontendEngineProps extends IFrontendEngineProps {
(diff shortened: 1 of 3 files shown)


ryhuz (dismissed):

weili-govtech:
technically this is not resetting the current location, it is more of confirming whether to refresh user's current location

in LocationModal's `getCurrentLocation()` ([source](https://github.com/LifeSG/web-frontend-engine/blob/main/src/components/fields/location-field/location-modal/location-modal.tsx#L283)), there is a similar `get-current-location` event, however I understand in this case, we don't even want to toggle `gettingCurrentLocation` to true. given the similarity in behaviour, may I suggest to:
- move the dispatch of event to `getCurrentLocation()` in LocationModal, above `setGettingCurrentLocation(true);`
- rename event to dispatch as `before-get-current-location`
- rename the second event as `refresh-current-location`

weili-govtech (changes requested):
technically this is not resetting the current location, it is more of confirming whether to refresh user's current location

in LocationModal's `getCurrentLocation()` ([source](https://github.com/LifeSG/web-frontend-engine/blob/main/src/components/fields/location-field/location-modal/location-modal.tsx#L283)), there is a similar `get-current-location` event, however I understand in this case, we don't even want to toggle `gettingCurrentLocation` to true. given the similarity in behaviour, may I suggest to:
- move the dispatch of event to `getCurrentLocation()` in LocationModal, above `setGettingCurrentLocation(true);`
- rename event to dispatch as `before-get-current-location`
- rename the second event as `refresh-current-location`

weili-govtech (approved):

PR #261, opened 2024-04-17
Mol 15398

zxianteo:
**Changes**
add multiple prop to allow user upload multiple images at once

-   [delete] branch

**Additional information**

-   You may refer to this [MOL-15398](https://jira.ship.gov.sg/browse/MOL-15398)
-   If user upload more than the maxFiles at the review modal, the modal will close and will show the exceed error


src/components/fields/image-upload/image-review/image-thumbnails/image-thumbnails.tsx
```diff
@@ -78,7 +80,7 @@ export const ImageThumbnails = (props: IProps) => {
 						<BorderOverlay isSelected={activeFileIndex === index} />
 					</ThumbnailItem>
 				);
-			} else if (image.addedFrom === "reviewModal") {
+			} else if (image.addedFrom === "reviewModal" || image.status < EImageStatus.NONE) {
```
weili-govtech, line 83: `image.status < EImageStatus.NONE` is not needed since there are conditions above that exclude other statuses already

src/components/fields/image-upload/image-review/image-review.tsx
```diff
@@ -117,30 +119,36 @@ export const ImageReview = (props: IProps) => {
...
+							addedFrom: "reviewModal",
+							slot,
+						},
+					];
+				});
 			});
+		} else {
+			setImages(images.filter(({ status }) => status == EImageStatus.UPLOADED));
+			setExceedError(true);
+			onExit();
```
weili-govtech, line 151: it's rather abrupt to close the review modal, perhaps we can have the review modal show an error message instead? (it's already showing error messages for unsupported photo formats or exceeded file size), may I trouble you to check with UX on this?

weili-govtech (changes requested):

src/components/fields/image-upload/image-review/image-thumbnails/image-thumbnails.tsx
zxianteo: but we want it to be show the file is error when we upload multiple image

weili-govtech: yes my point is there are conditions above that handle other statuses already, by the time the logic flow to this final if block, only the erroneous statuses are left, hence there is no need to check again

```
if (image.status === EImageStatus.NONE) { // cater for none status
	// ...
} else if (image.status > EImageStatus.NONE) { // cater for valid statuses
	// ...
} else if (image.addedFrom === "reviewModal") { // left with invalid statuses
	// ...
}
```

zxianteo: no. that error only will show when u added in reviewmodal, but if add at the input there will not show the error message in the review modal.

https://github.com/LifeSG/web-frontend-engine/assets/122598226/a4363a10-fca5-4093-a928-5fb0dbf4eb06

zxianteo: this is what we want

https://github.com/LifeSG/web-frontend-engine/assets/122598226/1fa6fa16-5107-481d-a627-888c4382ca10

weili-govtech: hello ok I get what you mean now then it's not an issue.

however I find it odd to see the error also appearing in the main image upload component just before the review confirmation modal appear. I think that error does not need to be shown, may I check if this has gone through ux?

src/components/fields/image-upload/image-upload.tsx
```diff
@@ -211,17 +212,20 @@ export const ImageUploadInner = (props: IGenericFieldProps<IImageUploadSchema>)
...
 	const renderImageReviewModal = () => {
+		const maxRule = validation?.find((rule) => "max" in rule);
 		return (
 			<ImageReview
 				accepts={acceptedFileTypes}
 				dimensions={dimensions}
 				id={id}
 				className={className}
 				maxFiles={maxFiles}
+				maxFilesErrorMessage={maxRule?.errorMessage || ERROR_MESSAGES.UPLOAD("photo").MAX_FILES(maxFiles)}
```
weili-govtech, line 223: may I suggest to put the fallback in `ImageError`? because `maxFilesErrorMessage` is an optional prop there and it is clearer to see the is a default message there

in addition, do use the `MAX_FILES_WITH_REMAINING` error message instead because user should still be able to upload some more images in the review modal.
ref: https://github.com/LifeSG/web-frontend-engine/blob/main/src/components/fields/image-upload/image-input/image-input.tsx#L156

src/components/fields/image-upload/image-review/image-error/image-error.tsx
```diff
@@ -10,6 +10,7 @@ const WARNING_ICON = "https://assets.life.gov.sg/web-frontend-engine/img/icons/w
 interface IProps extends Omit<ISharedImageProps, "maxFiles"> {
 	image: IImage;
 	onClickOk: (e: React.MouseEvent<HTMLButtonElement>) => void;
+	maxFilesErrorMessage?: string;
```
weili-govtech, line 13:
```suggestion
	maxFilesErrorMessage?: string | undefined;
```

src/components/fields/image-upload/image-review/image-review.tsx
```diff
@@ -61,6 +61,8 @@ interface IProps extends ISharedImageProps {
 	onExit: () => void;
 	outputType: string;
 	show: boolean;
+	multiple?: boolean | undefined;
+	maxFilesErrorMessage?: string;
```
weili-govtech, line 65:
```suggestion
	maxFilesErrorMessage?: string | undefined;
```

weili-govtech (changes requested):

weili-govtech:
may I trouble you to add test cases for the new behaviours?

src/__tests__/components/fields/image-upload/image-upload.spec.tsx
```diff
@@ -1041,4 +1041,68 @@ describe("image-upload", () => {
 			expect(input.getAttribute("capture")).toBe("user");
 		});
 	});
+
+	describe("upload multiple images at once", () => {
```
weili-govtech, line 1045:
```suggestion
	describe("upload multiple images at once through file dialog", () => {
```

weili-govtech (approved):

PR #267, opened 2024-04-25
Review update

weili-govtech:
**Changes**
- Updated accordion variant to use UneditableSection
- Added ability to unmask via api and error handling
- delete branch


src/components/custom/review/review.tsx
```diff
@@ -32,20 +39,139 @@ export const Review = (props: IGenericCustomElementProps<IReviewSchema>) => {
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, []);
 
+	useEffect(() => {
```
qroll, line 42: side note: potentially resets the failure count if the schema changes and causes the `schema.items` reference to change. tested with a static schema and didn't trigger the useEffect, so I don't think it's an actual issue

src/__tests__/components/custom/review/review.spec.tsx
```diff
@@ -111,49 +228,60 @@ describe(REFERENCE_KEY, () => {
...
+		it("should be able to render the bottomSection", () => {
+			renderComponent({
+				overrideField: {
+					variant,
+					description: DESCRIPTION,
+					bottomSection: {
+						alertBottom: {
+							uiType: "alert",
+							type: "warning",
+							children: ALERT_BOTTOM,
```
qroll, line 242: bottom section config is duplicated from the previous test case (lines 220-225)

src/components/custom/review/review.tsx
weili-govtech: just to be safe, I changed to use `useDeepCompareEffect` instead so `itemDetailList` will only be updated if the items have truly changed.

qroll (approved):

PR #282, opened 2024-05-24
Style HTML content

qroll:
**Changes**

- use `Markup` to style parsed HTML content using the sanitize component
- this supports theming for link color etc.
-  [delete] branch

**Changelog entry**

-   Apply default styling for HTML string children

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1360)

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The intention of frontend engine is to take out the heavy lifting of form develo
 
 Developers are expected to have the following packages installed:
 
--   @lifesg/react-design-system 2.5.0-canary.5
+-   @lifesg/react-design-system 2.6.0-canary.2
 -   @lifesg/react-icons 1.2.0
 -   react 17.0.2 or 18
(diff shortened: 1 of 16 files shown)


weili-govtech:
Sanitize instance [here](https://github.com/LifeSG/web-frontend-engine/blob/main/src/components/fields/image-upload/image-input/image-input.tsx#L211) requires the inline prop otherwise it will be rendered as a div within p.

qroll:
> Sanitize instance [here](https://github.com/LifeSG/web-frontend-engine/blob/main/src/components/fields/image-upload/image-input/image-input.tsx#L211) requires the inline prop otherwise it will be rendered as a div within p.

based on the example usage, it's similar to file-upload which can handle HTML flow content, so I'm proposing to render the content wrapper as a `div` instead of `p`

weili-govtech (approved):

PR #283, opened 2024-06-05
[MOL-15191] update include images with custom error status to uploadedImages;  add in uploadResponse when set image value 

TinYu2019:


src/components/fields/image-upload/image-upload.tsx
```diff
@@ -70,6 +77,7 @@ export const ImageUploadInner = (props: IGenericFieldProps<IImageUploadSchema>)
 					addedFrom: "schema",
 					uploadProgress: 100,
 					slot: i,
+					uploadResponse,
```
ryhuz, line 80: sorry i forgot why we need this

src/components/fields/image-upload/types.ts
```diff
@@ -90,6 +90,12 @@ export interface IUpdateImageStatus {
 	errorMessage?: string | undefined;
 }
 
+export interface IUploadedImage {
+	dataURL: string;
+	fileName: string;
+	uploadResponse: string;
```
weili-govtech, line 96: `uploadResponse` can be anything returned by the upload api

```suggestion
	uploadResponse?: any;
```

src/components/fields/image-upload/image-upload.tsx
weili-govtech: I think it's fine to add this, this can be used if a consumer wants to prefill with uploaded images + upload response for their own processing

weili-govtech (changes requested):
since the muted error state is rendered differently in the image review modal, can we have another story to demo it?

src/components/fields/image-upload/image-upload.tsx
TinYu2019: we need this when coming back to provide details page from edit page, otherwise the image cannot be loaded

src/components/fields/image-upload/types.ts
TinYu2019: updated

TinYu2019:
> since the muted error state is rendered differently in the image review modal, can we have another story to demo it?

we have one for the muted error already at here, do you mean this one? 
https://dev.designsystem.lifesg.io/web-frontend-engine/index.html?path=/story/field-imageupload-events--image-upload-ready-with-muted-error

weili-govtech (approved):

PR #284, opened 2024-06-06
Conditional validation fixes

weili-govtech:
**Changes**
- added generics to `IYupValidationRule` to inherit custom validation typings for `then` and `otherwise` rules
- fixed issue with non-conditional validation rules not running when there are conditional validation rules
- delete branch

**Changelog entry**
-   `then` and `otherwise` conditional validation typings now inherits custom validation typings
-   Ensure non-conditional validation rules can still work together with conditional validation rules

### DIFF
--- a/src/context-providers/yup/helper.ts
+++ b/src/context-providers/yup/helper.ts
@@ -200,13 +200,9 @@ export namespace YupHelper {
 					{
 						Object.keys(rule.when).forEach((fieldId) => {
 							const isRule = rule.when[fieldId].is;
-							const thenRule = mapRules(
-								mapSchemaType(yupSchema.type as TYupSchemaType),
-								rule.when[fieldId].then
-							);
+							const thenRule = mapRules(yupSchema, rule.when[fieldId].then);
 							const otherwiseRule =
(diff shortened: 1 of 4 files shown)


src/context-providers/yup/helper.ts
```diff
@@ -200,13 +200,9 @@ export namespace YupHelper {
...
 							const isRule = rule.when[fieldId].is;
-							const thenRule = mapRules(
-								mapSchemaType(yupSchema.type as TYupSchemaType),
-								rule.when[fieldId].then
-							);
+							const thenRule = mapRules(yupSchema, rule.when[fieldId].then);
 							const otherwiseRule =
-								rule.when[fieldId].otherwise &&
-								mapRules(mapSchemaType(yupSchema.type as TYupSchemaType), rule.when[fieldId].otherwise);
+								rule.when[fieldId].otherwise && mapRules(yupSchema, rule.when[fieldId].otherwise);
```
weili-govtech, line 205: this is the fix for non-conditional validation rules not running because it was taking a new Yup schema which loses all previous validation rules

qroll (approved):

PR #286, opened 2024-06-14
CCUBE-1389: Fix error not showing for nested field within radio button

weili-govtech:
**Changes**
-  Fixed error not showing for nested field within radio button
-  Replaced id check using `getNestedValueByKey` with `registeredFields` because the former cannot check nested arrays which nested fields are based on

**Changelog entry**

-   Ensure errors and warnings can be set for fields within radio buttons

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/CCUBE-1389)

### DIFF
--- a/src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
+++ b/src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
@@ -912,7 +912,7 @@ describe("frontend-engine", () => {
 			expect(getFieldOne().parentElement.nextSibling.textContent).toMatch(ERROR_MESSAGE);
 		});
 
-		it("should convert error object to string if the direct descendants don't match any fields", async () => {
+		it("should convert error object to string", async () => {
 			const handleClickNested = (ref: React.MutableRefObject<IFrontendEngineRef>) => {
 				ref.current.setErrors({
 					[FIELD_ONE_ID]: {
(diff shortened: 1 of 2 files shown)


qroll (approved):

qroll:
the test case for `should support setting of custom errors for nested fields` is failing

field1 is a wrapper and hence doesn't count as a registered field, so the child doesn't get evaluated

should it be called with `setError({ [FIELD_TWO]: "..."})` instead?

weili-govtech:
>  the test case for should support setting of custom errors for nested fields is failing
>
> field1 is a wrapper and hence doesn't count as a registered field, so the child doesn't get evaluated
>
> should it be called with setError({ [FIELD_TWO]: "..."}) instead? but since object is supported, would need to handle it still

alright I have consulted the users and it appears no one is writing in a nested manner, as such I have removed the ability to declare the error object in a nested manner and the test for it.

src/__tests__/components/frontend-engine/frontend-engine.spec.tsx
```diff
@@ -912,16 +912,6 @@ describe("frontend-engine", () => {
...
-			const handleClickNested = (ref: React.MutableRefObject<IFrontendEngineRef>) => {
-				ref.current.setErrors({ [FIELD_ONE_ID]: { [FIELD_TWO_ID]: ERROR_MESSAGE } });
-			};
-			render(<FrontendEngineWithCustomButton data={NESTED_JSON_SCHEMA} onClick={handleClickNested} />);
-			await waitFor(() => fireEvent.click(getCustomButton()));
-
-			expect(getFieldTwo().parentElement.nextSibling.textContent).toMatch(ERROR_MESSAGE);
-		});
-
 		it("should convert error object to string if the direct descendants don't match any fields", async () => {
```
qroll, line 915: this test case can be reworded as well since there's no concept for recursive matching any more

qroll (approved):

PR #288, opened 2024-06-19
[MOL-15191][SW] only handle default values for images  and files on f…

ryhuz:
`useEffect` to handle default values is supposed to only run on first load (if `is not dirty`)
however, if default value is exactly the same as the current form value, it is treated as not dirty even if there were new images added and removed.

solution is to add a new key to the value to determine if it has already been handled as default. (it also ensures that the handled value and the default value are never congruent)

* previous solution was to set `touched` status on adding new images (on top of the default) and only run the initial `useEffect` if not dirty and not touched. however, with the above fix, there is no need to check for `touched` status anymore. however, left in setting of `touched` just as 'correctness'

since file upload is implemented with the same logic, have applied the changes there too

**Changes**
Description of changes...

-   [delete] branch

### DIFF
--- a/src/components/fields/image-upload/image-upload.tsx
+++ b/src/components/fields/image-upload/image-upload.tsx
@@ -40,7 +40,6 @@ export const ImageUploadInner = (props: IGenericFieldProps<IImageUploadSchema>) 
 		},
 		id,
 		isDirty,
-		isTouched,
 		value,
 		...otherProps
 	} = props;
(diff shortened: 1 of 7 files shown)


src/components/fields/image-upload/image-upload.tsx
```diff
@@ -60,11 +61,11 @@ export const ImageUploadInner = (props: IGenericFieldProps<IImageUploadSchema>)
 	// =============================================================================
 	useEffect(() => {
 		dispatchFieldEvent("mount", id);
-	}, []);
+	}, [dispatchFieldEvent, id]);
 
 	useEffect(() => {
```
weili-govtech, line 66: add `isTouched` to the dependencies, do the same for `file-upload`

weili-govtech (changes requested):
can we have tests for the setting of the touched state?

src/components/fields/image-upload/image-upload.tsx
ryhuz: updated the dependency array

PR #291, opened 2024-07-01
[MOL-15190][SW] add a firstload flag to prevent retriggering get-sele…

ryhuz:
Fix bug where `recenterAndTriggerEvent` was running on form value updates

-   [delete] branch

### DIFF
--- a/src/components/fields/location-field/location-modal/location-modal.tsx
+++ b/src/components/fields/location-field/location-modal/location-modal.tsx
@@ -1,6 +1,6 @@
 import { MediaWidths, Modal } from "@lifesg/react-design-system";
 import { isEmpty } from "lodash";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


src/components/fields/location-field/location-modal/location-modal.tsx
```diff
@@ -313,7 +315,10 @@ const LocationModal = ({
 				});
 			}
 		};
-		recenterAndTriggerEvent();
+		if (isFirstOpen.current) {
+			recenterAndTriggerEvent();
+		}
+		isFirstOpen.current = false;
```
weili-govtech, line 321: feels slightly to start off with a truthy value, can we do the reverse?
```ts
const hasShownBefore = useRef(false);
// ...
useEffect(() => {
  // ...
  if (!hasShownBefore.current) {
	hasShownBefore.current = true;
	recenterAndTriggerEvent();
  }
}, [/*...*/]);
```

weili-govtech (approved):

PR #292, opened 2024-07-02
E-signature field

weili-govtech:
**Changes**
- New e-signature field based on design system component
- delete branch


src/components/fields/e-signature-field/e-signature-field.tsx
```diff
@@ -0,0 +1,109 @@
...
+			onChange({ target: { value: { fileId, dataURL: signatureDataURL } } });
+		}
+	};
+
+	// =============================================================================
+	// HELPER FUNCTIONS
+	// =============================================================================
+	const uploadFile = async (fileId: string, signatureDataURL: string) => {
+		setLoadingProgress(0);
+		const formData = new FormData();
```
qroll, line 68: just noting here that drafts `attachments.multipart` endpoint was based on file upload and expects `sessionId` and `slot`, we'll need to update these fields to be optional

qroll (changes requested):

src/components/fields/e-signature-field/e-signature-field.tsx
weili-govtech: noted, actually I can add `sessionId` if it's needed but I am looking to do away with `slot`

qroll: ok to leave these two fields out, they're actually optional on the db level anyway

qroll (approved):

PR #293, opened 2024-07-02
Ensure custom schema typing is still inherited when nested in other schemas

weili-govtech:
**Changes**
- when custom schema interface is passed as a generic to frontend engine, ensure the typing is inherited in all other schemas
-   delete branch

Example
```ts
interface MyCustomSchema extends TCustomComponentSchema<"my-custom-component", MyCustomRule> {
	myCustomKey: string;
}

const JSON_SCHEMA: IFrontendEngineData<undefined, MyCustomSchema> = {
	sections: {
		section: {
			uiType: "section",
			children: {
				wrapper: {
					uiType: "div",
					children: {
						custom: {
							referenceKey: "my-custom-component", // <-- this should be supported
							myCustomKey: "hello", // <-- this should be supported
						},
					},
				}
			},
		},
	},
};```

### DIFF
--- a/src/components/custom/filter/filter-item/types.ts
+++ b/src/components/custom/filter/filter-item/types.ts
@@ -3,9 +3,9 @@ import { ICustomElementJsonSchema } from "../../types";
 import { TClearBehavior } from "../filter/types";
 import { IFilterItemLabel } from "../types";
 
-export interface IFilterItemSchema<V = undefined> extends ICustomElementJsonSchema<"filter-item"> {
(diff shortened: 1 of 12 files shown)


src/components/custom/filter/filter-item/types.ts
weili-govtech, line 1: for all the generics named in FEE:
- `V` refers to custom validation typing
- `C` refers to custom component typing

nobeltws: 👍

nobeltws (approved):
looks good!

PR #294, opened 2024-07-08
Remove `handledFromDefault` from file and image upload field values

qroll:
**Changes**

- The `handledFromDefault` flag is set in the field value, so it can leak across sessions and cause the file to be filtered out

```
// it's equivalent to doing this and the file is not reflected
DefaultValue.args = {
	...COMMON_STORY_ARGS,
	defaultValues: [
		{
			fileId: "file-1",
			fileName: "test.jpg",
			dataURL: jpgDataURL,
			handledFromDefault: true,
		},
	],
};
```

- Had to find an alternative approach:
  - First, using `isDirty` and `isTouched` to determine if default values should be handled
  - Second, detect value changes. So the current uploaded files still need to be tracked
- Some issues with this
  - Need a way to uniquely identify files - the file id seems to do the job (`generateRandomId` should be good enough?)
  - `image-upload` does not track the id - have to generate for older fields
- [delete] branch

### DIFF
--- a/src/components/fields/image-upload/image-manager/image-manager.ts
+++ b/src/components/fields/image-upload/image-manager/image-manager.ts
@@ -33,7 +33,7 @@ export const ImageManager = (props: IProps) => {
 	// CONST, STATE, REFS
 	// =============================================================================
 	const { accepts, compress, dimensions, editImage, id, maxSizeInKb, outputType, upload, value } = props;
-	const { images, setImages, setErrorCount } = useContext(ImageContext);
+	const { images, setImages, setErrorCount, setCurrentFileIds } = useContext(ImageContext);
 	const previousImages = usePrevious(images);
 	const previousValue = usePrevious(value);
 	const { setValue } = useFormContext();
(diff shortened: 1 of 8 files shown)


ryhuz:
re: `handledFromDefault`
i tried this solution but because i also wanted to properly populate the `useEffects`, it was triggering infinite re-renders. haven't thought of a solution other than removing `values` from the array

src/components/fields/image-upload/image-manager/image-manager.ts
```diff
@@ -186,13 +186,15 @@ export const ImageManager = (props: IProps) => {
 		const hasNotPrefilledImages = notPrefilledImages.length > 0;
 		const shouldDirty = hasNotPrefilledImages || gotDeleteImages;
 
+		setCurrentFileIds(uploadedImages.map(({ id }) => id));
+
 		setValue(
 			id,
-			uploadedImages.map(({ dataURL, drawingDataURL, name, uploadResponse, addedFrom }) => ({
+			uploadedImages.map(({ id, dataURL, drawingDataURL, name, uploadResponse }) => ({
+				fileId: id,
```
weili-govtech, line 194: @ryhuz just want to be safe, this doesnt break OS logic right? I have checked the submit case logic and see that you guys build the payload manually so it's fine but I am not sure if other parts of the application will be affected

ryhuz: have discussed with ruoling and tested this branch with OS implementation. works ok.

if any other problem presents, will take it separately

ryhuz (approved):

weili-govtech (approved):

PR #301, opened 2024-07-18
Introduce typings for event listeners

weili-govtech:
**Changes**
- Introduce typings for `addFieldEventListener`, `dispatchFieldEvent` and `removeFieldEventListener`
- This is done via event overloading
- Note: This is just syntactic sugar to ease usage of the event listener methods and these new signatures are not used internally

```
// old way (still supported)
formRef.current.addFieldEventListener("click", "fieldId", myEventListener);
formRef.current.dispatchFieldEvent("update-image-status", "fieldId", myPayload);
formRef.current.removeFieldEventListener("click", "fieldId", myEventListener);

// new way with uiType as first argument
formRef.current.addFieldEventListener("button", "click", "fieldId", myEventListener);
formRef.current.dispatchFieldEvent("image-upload", "update-image-status", "fieldId", myPayload);
formRef.current.removeFieldEventListener("button", "click", "fieldId", myEventListener);
```


src/components/fields/location-field/types.ts
qroll, line 1: there seems to be a field event for `set-current-location` (which is different from the trigger?)

qroll (commented):
having autocompletion functionality is great 👍

src/components/fields/location-field/types.ts
weili-govtech: yeah this implementation is rather confusing, it is a mean for various components to communicate with each other. the component needs major refactoring in the future though I am not sure when we can make it happen

qroll: meaning `set-current-location` event is internal and doesn't need to be included in the public typings?

weili-govtech: yes it's internal and not meant to be called externally, hence not exposed in the public typings

PR #302, opened 2024-07-22
Render toggle nested fields in composite section

qroll:
**Changes**

Renders the nested field for checkbox and radio options in the composite section instead of sublabel

However, this is a **breaking change** for styling as the new layout requires the consumer to apply the appropriate padding around the content

-   [delete] branch

### DIFF
--- a/src/components/fields/checkbox-group/checkbox-group.tsx
+++ b/src/components/fields/checkbox-group/checkbox-group.tsx
@@ -9,14 +9,7 @@ import { TestHelper } from "../../../utils";
 import { useValidationConfig } from "../../../utils/hooks";
 import { Wrapper } from "../../elements/wrapper";
 import { ERROR_MESSAGES, Sanitize, Warning } from "../../shared";
-import {
-	CheckboxContainer,
-	Label,
-	StyledCheckbox,
-	StyledToggle,
-	ToggleSublabel,
-	ToggleWrapper,
-} from "./checkbox-group.styles";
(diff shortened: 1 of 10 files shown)


src/components/fields/checkbox-group/checkbox-group.tsx
```diff
@@ -158,12 +151,13 @@ export const CheckboxGroup = (props: IGenericFieldProps<TCheckboxGroupSchema>) =
...
-											<Wrapper>{option.children}</Wrapper>
-										</ToggleSublabel>
-									) : null
+								compositeSection={
+									option.children
+										? {
+												children: <Wrapper>{option.children}</Wrapper>,
+												collapsible: false,
+										  }
+										: null
```
weili-govtech, line 160: help me disable the indent linting rule for this component and the radio button

```suggestion
									option.children
										? /* eslint-disable indent */
										  {
												children: <Wrapper>{option.children}</Wrapper>,
												collapsible: false,
										  }
										: /* eslint-enable indent */
										  null
```

weili-govtech (changes requested):

src/components/fields/checkbox-group/checkbox-group.tsx
qroll: since prettier and eslint conflict, could the indent rule be disabled? the linter still flags out any formatting mismatch

weili-govtech: yeah that's fine too

weili-govtech (approved):

PR #307, opened 2024-07-26
Support icon or label add-on in text field

qroll:
**Changes**

Uses [DS InputGroup](https://designsystem.life.gov.sg/react/index.html?path=/docs/form-inputgroup--docs) to display add ons

-   [delete] branch

**Changelog entry**

-   Allow display of icon or label prefix/suffix in `text-field`

### DIFF
--- a/src/components/fields/text-field/types.ts
+++ b/src/components/fields/text-field/types.ts
@@ -7,6 +7,7 @@ import { IBaseFieldSchema } from "../types";
 type TCustomOptions = {
 	preventCopyAndPaste?: boolean | undefined;
 	preventDragAndDrop?: boolean | undefined;
+	addOn?: TIconAddOn | TLabelAddOn | undefined;
 };
 
 type TIconAddOn = {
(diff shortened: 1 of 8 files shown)


src/components/fields/text-field/types.ts
```diff
@@ -7,7 +9,21 @@ type TCustomOptions = {
...
+};
+
+type TLabelAddOn = {
+	type: "label";
+	value: string;
+	position?: InputGroupAddonPosition | undefined;
+};
+
 export type TCustomOptionsText = TCustomOptions & {
+	addOn?: TIconAddOn | TLabelAddOn | undefined;
```
weili-govtech, line 26: why not add to all text fields, i.e. add to `TCustomOptions`?

weili-govtech (changes requested):

src/components/fields/text-field/types.ts
qroll: at first I didn't think label addons made sense for the email field. but actually the icon is still useful, plus the numeric field can use both too. thanks for pointing this out!

weili-govtech (approved):

PR #308, opened 2024-07-29
Update design system

qroll:
**Changes**

Bump DS peer dependency

Notable changes in `v2.7.0-canary.3`:
- select and multi-select have a new dropdown menu
- the `listStyleWidth` prop is deprecated - have removed the stories
- list items now have aria role `option` instead of being buttons - update selectors in unit tests
---
-   [delete] branch

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
 		"yup": "^0.32.11"
 	},
 	"peerDependencies": {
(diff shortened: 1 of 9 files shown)


package.json
weili-govtech, line 1: need to update `peerDependencies` too

weili-govtech (changes requested):

weili-govtech (approved):

PR #311, opened 2024-08-06
[TIMELINE][WL] add timeline component

weili-govtech:
**Changes**
- add timeline component
- timeline item supports plain text, HTML text and FEE schema

### DIFF
--- a/src/components/custom/timeline/timeline.tsx
+++ b/src/components/custom/timeline/timeline.tsx
@@ -1,0 +1,70 @@
+import { Timeline as DSTimeline } from "@lifesg/react-design-system/timeline";
+import * as Icons from "@lifesg/react-icons";
+import { useEffect } from "react";
+import * as Yup from "yup";
+import { useValidationConfig } from "../../../utils/hooks";
+import { Sanitize } from "../../shared";
+import { IGenericCustomElementProps } from "../types";
+import { ITimelineSchema } from "./types";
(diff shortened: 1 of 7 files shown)


src/components/custom/timeline/timeline.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+	// =========================================================================
+	// HELPER FUNCTIONS
+	// =========================================================================
+	const buildItems = () => {
+		return items.map(({ children, label, statuses, ...restOfItem }) => {
+			let itemContent: React.ReactNode;
+			if (typeof children === "string") {
+				itemContent = (
+					<p>
```
weili-govtech, line 41: found that without `p`, the font size will become 1rem

qroll: oops, it's probably inheriting the styles as it's not defined by either FEE Sanitize or DS Timeline

the 18px is coming from either storybook's or the browser's default?
Screenshot 2024-08-07 at 10 51 10 AM

so it could potentially inherit another set of styles in a different context

would propose to apply on the markup helper in the DS Timeline

weili-govtech: oh right, it's caused by sgds.css, I will be removing it in another PR
actually it is already rendering via `Markup` component through the `Sanitize` component
anyway I decided to DS to implement `Text.Body` for the string-based text

qroll (approved):

PR #320, opened 2024-09-05
Create popover element

qroll:
**Changes**
Add popover element that renders an interactive text string and supports an optional icon based on the DS collection

It can be used standalone (with a parent container) or inline with text elements (see `Inline Usage` story)

The sanitisation behaviour in the text component was tweaked to prevent stripping of the inline svg icon. As a side effect, it seems to restore the paragraph styling

Screenshot 2024-09-05 at 11 08 19 AM

-   [delete] branch

**Changelog entry**

-   Add popover element

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-16147)


src/__tests__/components/elements/popover/popover.spec.tsx
weili-govtech, line 1: add another test for the rendering of the icon

weili-govtech (changes requested):

weili-govtech (approved):

PR #321, opened 2024-09-09
New array-field to support dynamic lists

qroll:
**Changes**

- New `array-field` component
  - Accepts a field schema which is rendered in another FEE instance
  - Supports basic copy customisation
- Based on Sequential's implementation, with some enhancements to validation of child FEE instances
  - Need to detect reset event
  - Need to trigger validation on submit
  - Need to get the correct validation status even before user makes any input
- Tweak `Prompt` styling so that it is scrollable when the content extends beyond screen height
-   [delete] branch

**Changelog entry**

-   Add `array-field` field that supports dynamic lists where items can be added or removed

**Additional information**

-   You may refer to this [ticket](https://jira.ship.gov.sg/browse/MOL-16572)

### DIFF
--- a/src/__tests__/components/fields/array-field/array-field.spec.tsx
+++ b/src/__tests__/components/fields/array-field/array-field.spec.tsx
@@ -1,0 +1,382 @@
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import cloneDeep from "lodash/cloneDeep";
+import merge from "lodash/merge";
+import { FrontendEngine } from "../../../../components";
+import { IArrayFieldSchema } from "../../../../components/custom";
+import { IFrontendEngineData, IFrontendEngineRef } from "../../../../components/frontend-engine";
+import {
+	ERROR_MESSAGE,
+	FRONTEND_ENGINE_ID,
+	FrontendEngineWithCustomButton,
+	TOverrideField,
+	TOverrideSchema,
+	getErrorMessage,
+	getResetButton,
+	getResetButtonProps,
+	getSubmitButton,
+	getSubmitButtonProps,
+} from "../../../common";
+import { warningTestSuite } from "../../../common/tests/warnings";
+
+const SUBMIT_FN = jest.fn();
+const COMPONENT_ID = "field";
+const UI_TYPE = "array-field";
+const TEXT_FIELD_ID = "input";
+const TEXT_FIELD_LABEL = "TextField";
+
+const JSON_SCHEMA: IFrontendEngineData = {
+	id: FRONTEND_ENGINE_ID,
+	sections: {
+		section: {
+			uiType: "section",
+			children: {
+				[COMPONENT_ID]: {
+					referenceKey: UI_TYPE,
+					sectionTitle: "Section title",
+					fieldSchema: {
+						[TEXT_FIELD_ID]: {
+							uiType: "text-field",
+							label: TEXT_FIELD_LABEL,
+						},
+					},
+				},
+				...getSubmitButtonProps(),
+				...getResetButtonProps(),
+			},
+		},
+	},
+};
+
+const renderComponent = (overrideField?: TOverrideField<IArrayFieldSchema>, overrideSchema?: TOverrideSchema) => {
+	const json: IFrontendEngineData = merge(cloneDeep(JSON_SCHEMA), overrideSchema);
+	merge(json, {
+		sections: {
+			section: {
+				children: {
+					[COMPONENT_ID]: overrideField,
+				},
+			},
+		},
+	});
+	return render(<FrontendEngine data={json} onSubmit={SUBMIT_FN} />);
+};
+
+const getTextField = (index: number): HTMLElement => {
+	const els = screen.getAllByRole("textbox");
+	return els[index] ?? null;
+};
+
+const getAddButton = (): HTMLElement => {
+	return screen.queryByRole("button", { name: "Add" });
+};
+
+const getRemoveButton = (index: number): HTMLElement => {
+	const els = screen.queryAllByRole("button", { name: "Remove" });
+	return els[index] ?? null;
+};
+
+describe(UI_TYPE, () => {
+	beforeEach(() => {
+		jest.restoreAllMocks();
+	});
+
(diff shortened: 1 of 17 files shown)


qroll:
putting this in draft first as there are UI changes expected. however there's no change in the core logic

qroll:
noticed this issue when using the default `checkbox`. it works fine with toggle

https://github.com/user-attachments/assets/3e1b35a9-ca2e-40b8-afb0-672fec323086

would it be okay to update the id generation to be unique?
```
	const formatId = (index: number): string => {
		const unique = generateRandomId();
		return `${id}-${unique}`;
	};
```

src/__tests__/components/fields/array-field/array-field.spec.tsx
weili-govtech, line 1: add tests for `length` rule

weili-govtech (changes requested):

weili-govtech:
> noticed this issue when using the default `checkbox`. it works fine with toggle
> 
>  Screen.Recording.2024-09-13.at.6.36.57.PM.mov 
> would it be okay to update the id generation to be unique?
> 
> ```
> 	const formatId = (index: number): string => {
> 		const unique = generateRandomId();
> 		return `${id}-${unique}`;
> 	};
> ```

yes it's fine, just to be sure, you are referring to the id assigned to the `StyledCheckbox` right?
anyway this fix is required for radio-button as well

src/components/custom/array-field/array-field.tsx
```diff
@@ -0,0 +1,254 @@
...
+		setShowRemovePrompt(false);
+	};
+
+	// =============================================================================
+	// HELPERS
+	// =============================================================================
+	const initValue = (len: number) => {
+		return Array(len)
+			.fill(null)
+			.map(() => ({}));
```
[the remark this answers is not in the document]
qroll: using map here to ensure the object reference is unique for each element

src/components/custom/array-field/array-field.tsx
```diff
@@ -0,0 +1,254 @@
...
+		const length = lengthRule?.["length"] ?? undefined;
+		setMin(min ?? length);
+		setMax(max ?? length);
+		setLength(length);
+
+		setFieldValidationConfig(
+			id,
+			Yup.array()
+				.test("is-array-valid", validRule?.errorMessage || ERROR_MESSAGES.ARRAY_FIELD.INVALID, () => {
+					if (!Array.isArray(value) || !validRule?.["valid"]) return true;
```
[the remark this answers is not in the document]
qroll: - `Array.isArray(undefined)` would also be false
- meaning we can skip the check for existence of `validRule`, correct?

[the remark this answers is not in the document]
weili-govtech: > Array.isArray(undefined) would also be false, so this check would still hold?
yeah `undefined` yields false indeed, sorry I didnt explain properly, I think this should fail the validation. now I can submit without filling anything even though `ArrayFieldElement` is complaining I didnt fill in the name.

> meaning we can skip the check for existence of validRule, correct?
yah correct

src/__tests__/components/fields/array-field/array-field.spec.tsx
weili-govtech: can we also test that 2 sections will be added if length of 2 is specified?

weili-govtech: oops I think you missed out this comment

src/components/custom/array-field/array-field-element.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+		if (isEqual(formValues, sectionValues)) {
+			return;
+		}
+
+		if (formValues) {
+			Object.entries(formValues).forEach(([key, value]) => {
+				formRef.current?.setValue(key, value);
+			});
+		}
```
weili-govtech, line 48: found an issue with empty `formValues`
1. add a new section
2. key in value in 1st section and leave 2nd section empty
3. delete 1st section
4. the values from 1st section got transferred to the 2nd because it is not able to clear it

we either need to handle empty `formValues` or force a remount

src/components/custom/array-field/types.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export interface IArrayFieldRemoveConfirmationModal {
+	title?: string | undefined;
+}
+
+export interface IArrayFieldSchema<V = undefined>
+	extends IBaseCustomFieldSchema<"array-field", V, IArrayFieldValidationRule> {
+	// TODO: introduce unique rule for children of fieldSchema
+	fieldSchema: Record<string, TFrontendEngineFieldSchema>;
```
weili-govtech, line 22: need to inherit custom validation rules
```suggestion
	fieldSchema: Record<string, TFrontendEngineFieldSchema<V>>;
```

qroll: quick check, I recall that custom rules added through `addCustomValidation` are tied to a specific FEE instance based on the `yupId`. do they get inherited by a nested FEE instance?

weili-govtech: good point, I dont think it does because the `yupId` will be unique. there is a need to align the `yupId` in nested instances, not sure if it's possible to check for a parent context or find a way to pass to the nested FEE instance

src/components/custom/array-field/array-field-element.tsx
qroll: I wasn't able to clear the form using react-hook-form `reset()` due to the use of default values. I also didn't want to replicate the code in `reset-button` to manually iterate through the keys

instead, I added a unique key identifier for each section to prevent React from unmounting. this helps to preserve the state in each section and prevents unnecessary updates of `formValues` prop

src/components/custom/array-field/types.ts
qroll: it should be possible, I'm thinking either
- FEE to accept `yupId` as prop
- FEE to internally track its parent(s) and do a mapping. this one will be more complex

would propose to tackle this when the `unique` rule is added?

weili-govtech (approved):

PR #324, opened 2024-09-12
[NUMERIC][WL] update numeric-field to deal with values as numbers

weili-govtech:
**Changes**
- updated numeric-field to deal with values as numbers instead of strings
- delete branch

**Changelog entry**

-  Fixed `numeric-field` to submit values as numbers

**Additional information**

-   You may refer to this [MOL-16566](https://jira.ship.gov.sg/browse/MOL-16566)
-   There was a bug reported that `numeric-field` cannot be submitted because the value is a string and not a number and that failed our backend validation
Screenshot 2024-09-12 at 9 36 39 AM

### DIFF
--- a/src/components/fields/text-field/text-field.tsx
+++ b/src/components/fields/text-field/text-field.tsx
@@ -111,7 +111,8 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
 	// =============================================================================
 	const handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
 		if (uiType === "numeric-field") {
-			onChange({ target: { value: +event.target.value ?? undefined } });
+			const isNumber = !isNaN(parseFloat(event.target.value));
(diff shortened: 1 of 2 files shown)


src/components/fields/text-field/text-field.tsx
```diff
@@ -110,8 +110,8 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
 	// EVENT HANDLERS
 	// =============================================================================
 	const handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
-		if (uiType === "numeric-field" && !event.target.value) {
-			onChange({ target: { value: undefined } });
+		if (uiType === "numeric-field") {
+			onChange({ target: { value: +event.target.value ?? undefined } });
```
qroll, line 114: `+""` becomes 0 which may not be intended. we may still need the falsy conditon to set undefined, but input to be cast to number

PR #328, opened 2024-09-19
Add error-field to force invalid form

qroll:
**Changes**

- An invisible field that when present, forces the form to be invalid
- Useful for validating a finite set of input combinations on the frontend without having to create a custom rule
- Can be used together with `shown` rule to display an actual error message to the user. The default story has an example usage
- Exposes an `error` rule to customise the message, but this is used by `validation-schema-generator` only
-   [delete] branch

**Changelog entry**

-   Add `error-field` to force invalid form when shown

### DIFF
--- a/src/__tests__/components/fields/error-field/error-field.spec.tsx
+++ b/src/__tests__/components/fields/error-field/error-field.spec.tsx
@@ -1,0 +1,93 @@
+import { fireEvent, render, waitFor } from "@testing-library/react";
+import cloneDeep from "lodash/cloneDeep";
+import merge from "lodash/merge";
+import { FrontendEngine } from "../../../../components";
+import { IErrorFieldSchema } from "../../../../components/fields";
+import { IFrontendEngineData } from "../../../../components/types";
+import {
+	FRONTEND_ENGINE_ID,
+	TOverrideSchema,
+	getField,
+	getResetButtonProps,
+	getSubmitButton,
+	getSubmitButtonProps,
+} from "../../../common";
+
+const SUBMIT_FN = jest.fn();
+const COMPONENT_ID = "field";
+const PARENT_ID = "input";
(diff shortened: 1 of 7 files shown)


src/components/fields/error-field/types.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface IErrorFieldValidationRule extends IYupValidationRule {
+	/** for customising error message when field is present, not visible to user */
+	error?: boolean | undefined;
+}
+
+export interface IErrorFieldSchema
+	extends Pick<
+		IBaseFieldSchema<"error-field", undefined, IErrorFieldValidationRule>,
+		"showIf" | "uiType" | "validation"
+	> {}
```
ngweili, line 13: this field is special, all other validations don't have effect here, I propose we narrow the `validation` typing further to allow the `error` rule only

```suggestion
export interface IErrorFieldValidationRule {
	/** for customising error message when field is present, not visible to user */
	error: boolean;
	errorMessage: string;
}

export interface IErrorFieldSchema extends Pick<IBaseFieldSchema<"error-field">, "showIf" | "uiType"> {
	validation?: [IErrorFieldValidationRule];
}
```

ngweili: can we also allow the field to have `children`? this way any messages can be rendered within it and we can save the effort to do conditionally rendering on 2 components

ngweili (changes requested):

src/__tests__/components/fields/error-field/error-field.spec.tsx
weili-govtech, line 1: can we test the rendering of the children too?

weili-govtech (changes requested):

weili-govtech (approved):

PR #331, opened 2024-09-25
Support schema label in checkbox and radio

qroll:
**Changes**

- 38d9b1451bd132552dd3e8cd9ca868532250683e
  - enhanced checkbox/radio option label to support schema. the use case is to support popover at any position 
  - the children are restricted to [phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content) which we can usually think of as "inline" elements (in terms of presentation). currently the inline elements in FEE are `popover` and `span`
  - to restrict the typing of children, split up `IWrapperSchema` and `TElementSchema` to be more discrete
- 6028ab9ba9629b4577d150ef51c17428819dfecf
  - `IWrapperSchema` renamed to `TWrapperSchema` as it's now a type, not an interface
- f666c0eb92a5dfaab92b53aeb2089680edfcaf56
  - updating existing components that also restrict typing of children, to reuse the new discrete schemas
-   [delete] branch

**Changelog entry**

-   Support schema label in checkbox and radio
- Deprecate `IWrapperSchema` and replace with `TWrapperSchema`

### DIFF
--- a/src/components/elements/alert/types.ts
+++ b/src/components/elements/alert/types.ts
@@ -1,15 +1,10 @@
 import { AlertProps, AlertType } from "@lifesg/react-design-system";
 import { TComponentOmitProps } from "../../frontend-engine";
-import type { IDividerSchema } from "../divider";
-import type { IOrderedListSchema, IUnorderedListSchema } from "../list";
-import type { ITextSchema } from "../text";
-import { IBaseElementSchema } from "../types";
+import { IBaseElementSchema, TBlockElementSchema, TInlineElementSchema } from "../types";
 
 export interface IAlertSchema<V = undefined, C = undefined>
 	extends IBaseElementSchema<"alert">,
 		TComponentOmitProps<AlertProps> {
 	type: AlertType;
-	children:
-		| React.ReactNode
-		| Record<string, IDividerSchema | IOrderedListSchema<V, C> | ITextSchema | IUnorderedListSchema<V, C>>;
+	children: React.ReactNode | Record<string, TBlockElementSchema<V, C> | TInlineElementSchema>;
 }
--- a/src/components/elements/alert/types.ts
+++ b/src/components/elements/alert/types.ts
@@ -2,9 +2,11 @@ import { AlertProps, AlertType } from "@lifesg/react-design-system";
 import { TComponentOmitProps } from "../../frontend-engine";
 import { IBaseElementSchema, TBlockElementSchema, TInlineElementSchema } from "../types";
 
+type TAlertChildren<V, C> = Exclude<TBlockElementSchema<V, C>, IAlertSchema> | TInlineElementSchema;
+
 export interface IAlertSchema<V = undefined, C = undefined>
 	extends IBaseElementSchema<"alert">,
 		TComponentOmitProps<AlertProps> {
 	type: AlertType;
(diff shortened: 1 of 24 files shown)


src/components/fields/error-field/types.ts
```diff
@@ -8,10 +7,8 @@ export interface IErrorFieldValidationRule {
 	errorMessage?: string | undefined;
 }
 
-type TErrorFieldChildren = IAlertSchema | ITextSchema | IWrapperSchema | IOrderedListSchema | IUnorderedListSchema;
-
 export interface IErrorFieldSchema
 	extends Pick<IBaseFieldSchema<"error-field", undefined, IErrorFieldValidationRule>, "showIf" | "uiType"> {
 	validation?: [IErrorFieldValidationRule] | undefined;
-	children?: Record<string, TErrorFieldChildren> | undefined;
+	children?: Record<string, TBlockElementSchema> | undefined;
```
weili-govtech, line 13: can we exclude `IGridSchema` and `TWrapperSchema` but add `TInlineElementSchema`?

src/components/elements/alert/types.ts
```diff
@@ -1,15 +1,10 @@
...
+import { IBaseElementSchema, TBlockElementSchema, TInlineElementSchema } from "../types";
 
 export interface IAlertSchema<V = undefined, C = undefined>
 	extends IBaseElementSchema<"alert">,
 		TComponentOmitProps<AlertProps> {
 	type: AlertType;
-	children:
-		| React.ReactNode
-		| Record<string, IDividerSchema | IOrderedListSchema<V, C> | ITextSchema | IUnorderedListSchema<V, C>>;
+	children: React.ReactNode | Record<string, TBlockElementSchema<V, C> | TInlineElementSchema>;
```
weili-govtech, line 9: can we exclude
- `IAlertSchema`
- `IGridSchema<V, C>`
- `TWrapperSchema<V, C>`

weili-govtech (changes requested):

src/components/fields/error-field/types.ts
qroll: would grid be more appropriate as a container element then? I view it as a lower-level layout helper so it's the same usage as `div` for me

src/components/elements/alert/types.ts
qroll: missed out on alert 🙇 

for grid and div, same as above, was thinking along the lines of layout in the alert

src/components/fields/error-field/types.ts
weili-govtech: yeah it makes more sense but if we consider grid a container element, wrapper should be a container element as well. the implication is on the review component, it will have to be updated to accept the wrapper schema

src/components/elements/alert/types.ts
weili-govtech: got it, I will resolve this comment as the outcome for error-field typing will apply to alert typing

src/components/fields/error-field/types.ts
qroll: since wrapper is not clearly defined yet, I'll exclude it from the block elements and let components opt in

moving grid to container since it's based on the DS Layout, which is meant for top level content anyway

weili-govtech (approved):

PR #333, opened 2024-09-26
Fix prefilled option not populated for select field with overrides

qroll:
**Changes**

- caused by field mounting before the overrides are loaded
- this is because `Sections` gets its schema from props and overrides from the context. the context is delayed by 1 render since it depends on `useEffect` to populate
- update `Sections` to get schema from form context, this ensures it renders the current schema in sync with its children
-   [delete] branch

**Changelog entry**

-   Fix prefill not working for `select` and `multi-select` when using overridden options


weili-govtech (changes requested):
thanks for the fix, may I request to add the same test other fields that have values dependent on `options`? basically `checkbox`, `chips`, `multi-select`, `nested-multi-select`, `radio` and `range-select`?

weili-govtech (approved):

PR #343, opened 2024-10-16
Call onValueChange with latest validity when schema changes

qroll:
**Changes**

`onValueChange` is not triggered when schema changes e.g. due to conditional rendering. This can cause the validity status to be stale. Detect when the form validation changes and trigger this callback accordingly

-   [delete] branch

**Changelog entry**

-   [WARNING] Trigger onValueChange when schema changes

### DIFF
--- a/src/components/frontend-engine/frontend-engine.tsx
+++ b/src/components/frontend-engine/frontend-engine.tsx
@@ -1,6 +1,7 @@
 import { yupResolver } from "@hookform/resolvers/yup";
 import cloneDeep from "lodash/cloneDeep";
 import isEmpty from "lodash/isEmpty";
+import isEqual from "lodash/isEqual";
 import { ReactElement, Ref, forwardRef, useCallback, useEffect, useImperativeHandle, useRef, useState } from "react";
 import { FormProvider, useForm } from "react-hook-form";
 import useDeepCompareEffect, { useDeepCompareEffectNoCheck } from "use-deep-compare-effect";
@@ -18,6 +19,7 @@ import {
 	useFormSchema,
 	useFormValues,
 	useFrontendEngineForm,
+	usePrevious,
 	useValidationConfig,
 	useValidationSchema,
 } from "../../utils/hooks";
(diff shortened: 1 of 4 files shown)


src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -227,6 +229,11 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
 			const subscription = watch(() => {
 				onValueChange(getFormValues(undefined, stripUnknown), checkIsFormValid());
 			});
+
+			if (previousFormValidationConfig && previousFormValidationConfig !== formValidationConfig) {
```
weili-govtech, line 233: can leave a comment to explain why we need the manual call

weili-govtech: can we add another condition to fire onValueChange only if the form validity has changed? then we can also bring back the schema change unit test

weili-govtech (changes requested):

src/components/frontend-engine/use-form-change.ts
```diff
@@ -0,0 +1,154 @@
...
+			if (!isEqual(previousValidationRules, validationRules)) {
+				setHasSchemaChange(true);
+			}
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [formValidationConfig]);
+
+	useEffect(() => {
+		// re-evaluate form validity when schema changes
+		if (hasSchemaChange && previousHardValidationSchema !== hardValidationSchema) {
```
weili-govtech, line 117: I vaguely recall you talking about this, can explain again why the `previousHardValidationSchema !== hardValidationSchema` check? since both objects will never match each other

src/components/frontend-engine/use-form-change.ts
```diff
@@ -0,0 +1,154 @@
...
+					});
+				}
+
+				setOldFormValues(value);
+			});
+
+			return () => subscription.unsubscribe();
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [formState, watch]);
```
weili-govtech, line 151: why move this useEffect here? I feel that it has more to do with field errors than form changes

src/components/frontend-engine/use-form-change.ts
qroll: in the current behaviour, it just so happens `hasSchemaChange` and `hardValidationSchema` update in the same render cycle, but that may not be guaranteed in the future. it is safer to only trigger this when `hardValidationSchema` has definitely updated

src/components/frontend-engine/use-form-change.ts
qroll: the error handling is also in response to form changes via `watch`, so I thought it made sense to put them together. I can revert this if it's preferred

weili-govtech: I see, alright we keep this here then, we will move this out when there are enough error handling logic to warrant a separate file

weili-govtech (approved):

PR #349, opened 2024-10-24
[NL][Luna] extract datetime and gps from uploaded images

namlh023:
**Changes**

- Extract date time and GPS from uploaded images.
-   [delete] branch

**Additional information**

-   You may refer to this [MOL-15397](https://jira.ship.gov.sg/browse/MOL-15397)

### DIFF
--- a/src/components/fields/image-upload/image-manager/image-manager.ts
+++ b/src/components/fields/image-upload/image-manager/image-manager.ts
@@ -110,7 +110,6 @@ export const ImageManager = (props: IProps) => {
 								});
 								if (image.addedFrom !== "schema") {
 									compress ? compressImage(index, image) : convertImage(index, image);
-									extractImageMetadata(index, image);
 								}
 							} else {
 								setImages((prev) => {
@@ -235,11 +234,13 @@ export const ImageManager = (props: IProps) => {
 					return updatedImages;
 				});
 			} else {
+				const metadata = await ImageHelper.getMetadata(image.file);
 				setImages((prev) => {
 					const updatedImages = [...prev];
 					updatedImages[index] = {
 						...prev[index],
 						dataURL,
+						metadata,
 						status: EImageStatus.CONVERTED,
 					};
 					return updatedImages;
@@ -283,12 +284,14 @@ export const ImageManager = (props: IProps) => {
 					return updatedImages;
 				});
 			} else {
+				const metadata = await ImageHelper.getMetadata(imageToCompress.file);
 				const dataURL = await FileHelper.fileToDataUrl(compressed);
 				setImages((prev) => {
 					const updatedImages = [...prev];
 					updatedImages[index] = {
 						...prev[index],
 						dataURL,
+						metadata,
 						status: EImageStatus.COMPRESSED,
 					};
 					return updatedImages;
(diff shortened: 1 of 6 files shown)


src/components/fields/image-upload/image-manager/image-manager.ts
```diff
@@ -110,6 +110,7 @@ export const ImageManager = (props: IProps) => {
 								});
 								if (image.addedFrom !== "schema") {
 									compress ? compressImage(index, image) : convertImage(index, image);
+									extractImageMetadata(index, image);
```
ryhuz, line 113: can find a way to optimise this to avoid setting state twice for the same process?
now it's once in compress/convert image
once in extract data

ryhuz: that way you can also extract before compressing

src/components/fields/image-upload/image-manager/image-manager.ts
```diff
@@ -190,10 +191,11 @@ export const ImageManager = (props: IProps) => {
 
 		setValue(
 			id,
-			uploadedImages.map(({ id, dataURL, drawingDataURL, name, uploadResponse }) => ({
+			uploadedImages.map(({ id, dataURL, drawingDataURL, name, metadata, uploadResponse }) => ({
 				fileId: id,
 				fileName: name,
 				dataURL: drawingDataURL || dataURL,
+				metadata,
```
weili-govtech, line 198: can we have a new option in the schema to control whether to send the metadata? I want to avoid causing controversies by sending potentially sensitive data for other consumers of this component

src/components/fields/image-upload/types.ts
```diff
@@ -81,6 +81,7 @@ export interface IImage {
 	uploadResponse?: any;
 	slot: number;
 	customErrorMessage?: string | undefined;
+	metadata?: IImageMetadata;
```
weili-govtech, line 84:
```suggestion
	metadata?: IImageMetadata | undefined;
```

weili-govtech (changes requested):

weili-govtech:
may I confirm the extracting of geolocation metadata only works for specific scenarios? I just tried on android (15.0) chrome and it didnt work.

ryhuz:
> may I confirm the extracting of geolocation metadata only works for specific scenarios? I just tried on android (15.0) chrome and it didnt work.

this probably is due to the `input` having `accepts`. mentioned it to you the other day, but think we will look at a proper solution after this MR

src/components/fields/image-upload/image-manager/image-manager.ts
ryhuz: iinw this data can be extracted from the base64. so if that data is available to the consumer, then this metadata should not be any more sensitive.
@namlh023 is that correct?

namlh023: As long as we return the dataURL, the consumer can extract it to access the metadata. I'm just unsure about how much information they can retrieve from it.

namlh023: Can confirm that consumers can access all the image metadata from the data URL we provide (including GPS and DateTime). Therefore, adding a new option may not be necessary.

weili-govtech: ok my concern is that we are explicitly extracting the metadata now and putting them out in plain text as part of the submitted values.

however I think we can go ahead to do this first since the data is in the base64 payload in the first place, perhaps the question is whether we should be able to strip the metadata instead of not extracting the metadata. we will tackle this when there is a request to do it.

ryhuz:
no further comments about current changes. just pending test fix

src/components/fields/image-upload/types.ts
namlh023: updated

namlh023:
> no further comments about current changes. just pending test fix

I discovered that the core issue is that we didn't wait for the `convertImage` function to complete during the test. For this ticket, the solution is to mock the `getMetadata` function. However, in the future, if anyone calls an asynchronous function within `convertImage`, the tests may fail again.

ryhuz (approved):

weili-govtech (approved):

PR #366, opened 2024-12-10
Async validation

weili-govtech:
**Changes**
- Added support to perform async validation

**Additional information**
- This feature is to support validation via postMessage for the upcoming iframe field

### DIFF
--- a/src/__tests__/components/elements/wrapper/conditional-renderer.spec.tsx
+++ b/src/__tests__/components/elements/wrapper/conditional-renderer.spec.tsx
@@ -87,7 +87,7 @@ describe("conditional-renderer", () => {
 		${"email"}     | ${{ email: true }}           | ${"hello"} | ${"john@doe.tld"}
 		${"url"}       | ${{ url: true }}             | ${"hello"} | ${"https://domain.tld"}
 		${"uuid"}      | ${{ uuid: true }}            | ${"hello"} | ${"e9949c11-51b6-4c44-9070-623dfb2ca01a"}
-	`("should support $condition condition for string conditional rendering", ({ config, invalid, valid }) => {
+	`("should support $condition condition for string conditional rendering", async ({ config, invalid, valid }) => {
 		const uiType = "text-field";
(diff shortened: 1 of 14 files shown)


src/utils/hooks/use-validation-schema.ts
```diff
@@ -60,23 +60,25 @@ export const useValidationSchema = () => {
...
+					return acc;
+				}, {});
 
-			setWarnings(updatedWarnings);
-		}
+				setWarnings(updatedWarnings);
+			})
+			.then(() => {
+				setWarnings({});
+			});
```
qroll, line 81: this `then` would always run after the `catch`. wouldn't it end up replacing the warning state?

src/__tests__/components/elements/wrapper/conditional-renderer.spec.tsx
```diff
@@ -102,10 +102,10 @@ describe("conditional-renderer", () => {
 		};
 		renderComponent(fields);
 
-		fireEvent.change(getFieldOne(), { target: { value: invalid } });
+		await waitFor(() => fireEvent.change(getFieldOne(), { target: { value: invalid } }));
```
qroll, line 105: what is the purpose of `waitFor` for some of these events? would async `act` also work?

qroll (changes requested):

src/__tests__/components/elements/wrapper/conditional-renderer.spec.tsx
weili-govtech: the `waitFor` is added due to the change in `checkIsFormValid()` to be async, in some tests, the state is still updating after the tests have been completed.

I can use `act` as well, it will wait for the side effects to resolve though based on past experience, some times it does not wait long enough for unrelated side effects to resolve, hence I used `waitFor` here

qroll: I guess there's no harm using it specifically in these places since it's on a per action basis

qroll (approved):

PR #368, opened 2024-12-13
Fix textarea error and max length counter not being on the same line

jhyeep:
**Changes**
The error message and counter texts are not aligned. This PR changes `Textarea` to use `Form.Textarea` instead of the standalone component (form version has this fixed)
image

-   [delete] branch

**Changelog entry**
- Fixed textarea error message and max length counter not being on the same line

### DIFF
--- a/src/components/fields/textarea/textarea.tsx
+++ b/src/components/fields/textarea/textarea.tsx
@@ -97,23 +97,9 @@ export const Textarea = (props: IGenericFieldProps<ITextareaSchema>) => {
 		);
 	};
 
-	const renderLabel = () => {
-		const baseId = `${id}-base`;
-		if (typeof formattedLabel === "string") {
-			return <Form.Label htmlFor={baseId}>{formattedLabel}</Form.Label>;
-		} else {
-			return (
-				<Form.Label {...formattedLabel} htmlFor={baseId}>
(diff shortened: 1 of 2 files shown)


src/components/fields/textarea/textarea.tsx
```diff
@@ -96,15 +97,29 @@ export const Textarea = (props: IGenericFieldProps<ITextareaSchema>) => {
...
+				</Form.Label>
+			);
+		}
+	};
+
 	return (
 		<>
-			<Form.CustomField label={formattedLabel} id={id} errorMessage={error?.message}>
+			<Form.CustomField id={id}>
+				{renderLabel()}
```
qroll, line 116: what's the rationale for rendering label separately as well? would it be sufficient to remove the errorMessage prop but pass label to `Form.CustomField`?

qroll (changes requested):

src/components/fields/textarea/textarea.tsx
jhyeep: it was originally to make sure `htmlFor` points to the right id but i realized it might be unnecessary after all. test cases were also failing previously but they seem to work now, maybe i got misled by jest

qroll (approved):

PR #379, opened 2025-01-31
Set `hidden-field` value from schema

qroll:
**Changes**

- add `value` prop to `hidden-field`
- this will take precedence over form values and effectively force it to be uneditable
- for type safety, the typing has been updated so that `value` matches `valueType`. however this required converting the interface to a type
-   [delete] branch

**Changelog entry**

-   Set `hidden-field` value from schema to make the field uneditable

### DIFF
--- a/src/__tests__/components/fields/hidden-field/hidden-field.spec.tsx
+++ b/src/__tests__/components/fields/hidden-field/hidden-field.spec.tsx
@@ -122,6 +122,15 @@ describe(UI_TYPE, () => {
 		expect(SUBMIT_FN).toHaveBeenCalledWith(expect.objectContaining({ [COMPONENT_ID]: false }));
 	});
 
+	it("should set value to null if valueType=null", async () => {
+		const defaultValue = true;
+		renderComponent({ valueType: "null" }, { defaultValues: { [COMPONENT_ID]: defaultValue } });
+
+		await waitFor(() => fireEvent.click(getSubmitButton()));
+
+		expect(SUBMIT_FN).toHaveBeenCalledWith(expect.objectContaining({ [COMPONENT_ID]: null }));
+	});
+
 	it("should not let setValue override the schema value", async () => {
 		const json: IFrontendEngineData = merge(cloneDeep(JSON_SCHEMA), {
 			sections: {
@@ -216,6 +225,15 @@ describe(UI_TYPE, () => {
 			await waitFor(() => fireEvent.click(getSubmitButton()));
 
 			expect(SUBMIT_FN).toHaveBeenCalledWith(expect.objectContaining({ [COMPONENT_ID]: "hello" }));
+		});
+
+		it("should revert to null if valueType=null on reset", async () => {
+			renderComponent({ valueType: "null" }, { defaultValues: { [COMPONENT_ID]: "bye" } });
+
+			fireEvent.click(getResetButton());
+			await waitFor(() => fireEvent.click(getSubmitButton()));
+
+			expect(SUBMIT_FN).toHaveBeenCalledWith(expect.objectContaining({ [COMPONENT_ID]: null }));
 		});
 	});
 });
--- a/src/__tests__/components/fields/hidden-field/hidden-field.spec.tsx
+++ b/src/__tests__/components/fields/hidden-field/hidden-field.spec.tsx
@@ -105,6 +105,23 @@ describe(UI_TYPE, () => {
 		expect(SUBMIT_FN).not.toHaveBeenCalled();
 	});
 
+	it("should support setting of value from the schema", async () => {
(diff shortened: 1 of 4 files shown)


src/components/fields/hidden-field/types.ts
```diff
@@ -1,6 +1,38 @@
+import { IYupValidationRule } from "../../types";
 import { IBaseFieldSchema } from "../types";
 
-export interface IHiddenFieldSchema<V = undefined>
-	extends Pick<IBaseFieldSchema<"hidden-field", V>, "showIf" | "validation" | "uiType"> {
-	valueType?: "string" | "number" | "boolean" | undefined;
+export interface IHiddenFieldValidationRule extends IYupValidationRule {
+	/** for customising error message when submitted value does not match schema value */
+	matchesSchema?: boolean | undefined;
```
weili-govtech, line 6:
```suggestion
	equalsSchemaValue?: boolean | undefined;
```
I think we can keep to the `equals` prefix similar to `equals` and `equalsField` rules and be more specific because I was under the impression it is meant to match a schema structure.

other than that, why are we applying the validation in the validation-schema-generator only? the rule becomes useless if the validation-schema-generator is not used

src/components/fields/hidden-field/types.ts
```diff
@@ -1,6 +1,38 @@
...
+	valueType: "boolean";
+	value?: boolean;
+};
+
+type NoValueField = {
+	valueType?: never | undefined;
+	value?: never | undefined;
+};
+
+type FieldType = StringField | NumberField | BooleanField | NoValueField;
```
weili-govtech, line 29: - can prefix types with `T`
- add `undefined` union to optional properties

src/__tests__/components/fields/hidden-field/hidden-field.spec.tsx
```diff
@@ -105,6 +105,23 @@ describe(UI_TYPE, () => {
...
+
+	it("should not let default value override the schema value", async () => {
+		const defaultValue = true;
+		renderComponent({ valueType: "boolean", value: false }, { defaultValues: { [COMPONENT_ID]: defaultValue } });
+
+		await waitFor(() => fireEvent.click(getSubmitButton()));
+
+		expect(SUBMIT_FN).toHaveBeenCalledWith(expect.objectContaining({ [COMPONENT_ID]: false }));
+	});
+
```
weili-govtech, line 124: add another test for `setValue()` not to override the schema value

src/__tests__/components/fields/hidden-field/hidden-field.spec.tsx
```diff
@@ -147,5 +164,21 @@ describe(UI_TYPE, () => {
...
+						},
+					},
+				},
+			});
+			render(<FrontendEngineWithCustomButton data={json} onClick={handleClick} />);
+			fireEvent.click(screen.getByRole("button", { name: "Custom Button" }));
+
+			expect(formIsDirty).toBe(false);
+		});
 	});
```
weili-govtech, line 183: add another test to ensure the form is resetting to the schema value

weili-govtech (changes requested):

src/components/fields/hidden-field/types.ts
qroll: the hidden field always overrides the form value to match the schema value, hence I didn't expect any validation required on the frontend

weili-govtech: actually, under what circumstances will the submitted value differ from the schema value? shouldnt they always match? i.e. this rule is not necessary and should be inbuilt rather than declared by the consumer?

qroll: this is a safeguard for backend submission, in case the request body gets overridden

by inbuilt do you mean a generic error message? this rule is for anyone who may need to customise the error message on the backend

weili-govtech: I mean if `value` is declared in the schema, it should be implied that the submitted value will be equal to that value, hence any validation should be applied automatically (in validation-schema-generator). so the `equalsSchemaValue` rule is not necessary after all.

weili-govtech: oh right, I forgot that this is merely for customising the error message

src/components/fields/hidden-field/hidden-field.tsx
```diff
@@ -59,7 +64,7 @@ export const HiddenField = (props: IGenericFieldProps<THiddenFieldSchema>) => {
 			id={id}
 			data-testid={id}
 			name={name}
-			value={schemaValue ?? value}
+			value={schemaValue ?? value ?? ""}
```
qroll, line 67: prevents "controlled to uncontrolled" React warning

weili-govtech (approved):

PR #383, opened 2025-03-11
[MOL-17974][BX] fix to prevent the blur event when click pill

wilsontbx:
**Changes**
Description of changes...

-   [keep | delete] branch

**Changelog entry**

-   [BREAKING | WARNING | remove if none] Description of change

**Additional information**

-   You may refer to this [ticket](url)
-   Any other information you would like to include

### DIFF
--- a/src/components/fields/textarea/textarea.tsx
+++ b/src/components/fields/textarea/textarea.tsx
@@ -61,6 +61,12 @@ export const Textarea = (props: IGenericFieldProps<ITextareaSchema>) => {
 		onChange(event);
 	};
 
+	const onBlur = (event: React.FocusEvent<HTMLTextAreaElement>) => {
+		if (event.relatedTarget === null) {
+			event.target.focus();
+		}
+	};
+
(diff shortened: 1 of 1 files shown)


src/components/fields/textarea/textarea.tsx
```diff
@@ -73,6 +73,8 @@ export const Textarea = (props: IGenericFieldProps<ITextareaSchema>) => {
 				value: newValue,
 			},
 		});
+
+		trigger(id);
```
weili-govtech, line 77: the `onChange` call above should trigger validation already should this is redundant.

this bug appears to happen because user clicked away from the textarea and that triggered the blur event. in the unfortunate instance of the video shared, the effect of the validation happened after the text is appended.

I think a more complete fix will be to prevent the blur event when clicking on any of the pills, you can refer to https://adueck.github.io/blog/keep-focus-when-clicking-on-element-react/ but keep in mind to prevent the blur event ONLY if user clicks the pills and no other elements.

PR #385, opened 2025-03-13
[MOL-17974][MA] force focus on text area field when chips are clicked

mauw-paloit:
**Changes**

-   [delete] branch

**Changelog entry**

-   Add focus on text area when chips are clicked to fix bug with onBlur validation not firing

**Additional information**

-   You may refer to this ticket [MOL-17974](https://sgtechstack.atlassian.net/browse/MOL-17974)

### DIFF
--- a/src/components/fields/textarea/textarea.tsx
+++ b/src/components/fields/textarea/textarea.tsx
@@ -23,6 +23,7 @@ export const Textarea = (props: IGenericFieldProps<ITextareaSchema>) => {
 		schema: { className, chipTexts, chipPosition, rows = 1, resizable, label: _label, validation, ...otherSchema },
 		value,
 		warning,
+		onBlur,
 		...otherProps
 	} = props;
 	const { setValue } = useFormContext();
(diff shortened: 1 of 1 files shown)


src/components/fields/textarea/textarea.tsx
```diff
@@ -73,6 +74,10 @@ export const Textarea = (props: IGenericFieldProps<ITextareaSchema>) => {
 				value: newValue,
 			},
 		});
+
+		if (textAreaRef.current) {
+			textAreaRef.current.focus();
+		}
```
weili-govtech, line 80: this still results in a brief moment when the textarea gets blurred, the error message still gets displayed
I will recommend to re-focus the field based on: https://adueck.github.io/blog/keep-focus-when-clicking-on-element-react/
you can check if user clicked on the pills through `e.relatedTarget`


https://github.com/user-attachments/assets/8caf269c-5ea0-4d34-b687-d6350a577017

mauw-paloit: Thanks, updated the MR. I kept the original implementation of having a focus event when a chip is clicked to cater for the scenario that the user has triggered the validation message, clicked away, then clicks a pill. In this scenario, it should focus on the textArea element so if the user clicks away again, it is revalidated. To fix the issue you highlighted in your comment, I added a fix based on your suggestion to prevent onBlur from occurring if the user clicks a chip. Let me know your thoughts!

PR #388, opened 2025-03-28
Ccube 1703

weili-govtech:
**Changes**
- Merge to long-lived branch
- Initial upgrade to to DS v3
- Switch to deprecated v2 token
- Removal of deprecated `listStyleWidth` prop from `nested-multi-select`, other deprecated props to be removed in future PRs
- Temporarily disabling form builder story due to peer dependency conflict, form builder to be fixed when FEE v2 alpha build is released
- Note: Most of the changes are the switch to the v2 token, it will be easier to check by commits


qroll (changes requested):

qroll:
noticed that `npm ci` in the pipeline is failing due to clashing peer deps with form builder, but the tests are passing locally, so once form builder is fixed as you mentioned, all should be good

qroll (approved):

PR #391, opened 2025-04-09
[MISC][WL] temporarily remove form-builder

weili-govtech:
**Changes**
- Temporarily remove form-builder package for build to succeed


weili-govtech:
My bad, I didnt realise our build job was failing because of the form-builder conflicts

qroll (approved):

PR #401, opened 2025-05-26
Fix incorrect validation when rule is not available on Yup schema

qroll:
**Changes**

The Yup schema becomes undefined when trying to apply a custom rule that is not applicable to the schema type, this manifests as conditional rendering not working as expected

-   [delete] branch

Example story to test the behaviour

```
export const Example = Template.bind({});
Example.args = {
	field: {
		uiType: "hidden-field",
		valueType: "null",
		// valueType: "string", // uncomment to compare behaviour
		// value: "Error", // uncomment to compare behaviour
		validation: [
			{
				equalsSchemaValue: true,
				errorMessage: "Does not match",
			},
		],
	},
	error: {
		uiType: "error-field",
		showIf: [{ field: [{ filled: true }, { notMatches: "/^[a-z]+$/" }] }],
		children: {
			list: {
				uiType: "ordered-list",
				children: [
					"This is a list item",
					{
						errorMessage: {
							uiType: "list-item",
							showIf: [{ field: [{ filled: true }, { notMatches: "/^[a-z]+$/" }] }],
							children: {
								alert: {
									uiType: "text-body",
									children: "Contains invalid characters",
								},
							},
						},
					},
				],
			},
		},
	},
};
```

**Changelog entry**

-   Fix validation failure when rule is not available on Yup schema

### DIFF
--- a/src/context-providers/yup/helper.ts
+++ b/src/context-providers/yup/helper.ts
@@ -237,8 +237,8 @@ export namespace YupHelper {
 			const customRuleKey = Object.keys(rule).filter((k) =>
 				customYupConditions.includes(k as TYupCondition)
 			)?.[0] as TYupCondition;
-			if (customRuleKey) {
-				yupSchema = (yupSchema as unknown)[customRuleKey]?.(rule[customRuleKey], rule.errorMessage);
+			if (customRuleKey && (yupSchema as unknown)[customRuleKey]) {
+				yupSchema = (yupSchema as unknown)[customRuleKey](rule[customRuleKey], rule.errorMessage);
 			}
 			// prevent applying non-required validation for empty fields
 			if (yupSchema) {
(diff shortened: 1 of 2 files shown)


src/context-providers/yup/helper.ts
```diff
@@ -237,8 +237,8 @@ export namespace YupHelper {
 			const customRuleKey = Object.keys(rule).filter((k) =>
 				customYupConditions.includes(k as TYupCondition)
 			)?.[0] as TYupCondition;
-			if (customRuleKey) {
-				yupSchema = (yupSchema as unknown)[customRuleKey]?.(rule[customRuleKey], rule.errorMessage);
+			if (customRuleKey && (yupSchema as unknown)[customRuleKey]) {
+				yupSchema = (yupSchema as unknown)[customRuleKey](rule[customRuleKey], rule.errorMessage);
 			}
```
weili-govtech, line 242: can we follow the preceding codes and do `console.warn()` to inform the consumer that the rule is not applied?

```
console.warn(`error applying "${customRuleKey}" condition to ${yupSchema.type} schema`);
```

PR #423, opened 2025-07-17
[MOL-18941][TP] Fix issue image upload on some mobile devices

thanhpt0105-gt:
**Changes**
Add a small delay to **handleInputChange** in **drag-upload** so the files can be loaded properly.

-   delete branch


weili-govtech:
I need more context to the issue:
- understand it happens to certain phones, what is the underlying error?
- what was the root cause?
- is 100ms an arbitrary number? is it enough if bigger files are provided / slower phones are used?

weili-govtech:
is it possible to use `FileReader` instead of a `setTimeout`?

weili-govtech:
approving PR since other attempts did not work
- using `FileReader`
- using `requestAnimationFrame`

PR #433, opened 2025-07-31
[MOL-19058][MA] Add zoom duration to location field props

mauw-paloit:
**Changes**
Add `duration` to location field `mapPanZoom` prop. This allows us to control how long it takes for the map modal to zoom and center on a specified lat lng using the leaflet api

-   delete branch

**Changelog entry**

-   Add optional `duration` to location field `mapPanZoom` props

**Additional information**

### DIFF
--- a/src/components/fields/location-field/location-modal/location-picker/location-picker.tsx
+++ b/src/components/fields/location-field/location-modal/location-picker/location-picker.tsx
@@ -210,7 +210,7 @@ export const LocationPicker = ({
 				? map.getZoom()
 				: panZoomValue;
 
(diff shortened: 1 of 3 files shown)


src/components/fields/location-field/types.ts
weili-govtech, line 1: can we add as `duration` to `mapPanZoom`? I was under the impression the `zoomDuration` sets how fast I can zoom in / out

mauw-paloit: Thanks for reviewing. Updated!

weili-govtech (approved):

PR #436, opened 2025-08-18
Bump DS v3.0.0-alpha.19

qroll:
**Changes**

- General updates to align with latest design system version (excluding v3.0.0-alpha.20's React 19 upgrades and breaking changes)
  - Check for `aria-disabled` instead of `toBeDisabled()` for text inputs
  - Update selectors in unit tests
  - Ensure default `Button` is used when icon is not set in `array-field`, otherwise a `React.cloneElement` error is logged
-   [delete] branch


src/components/custom/array-field/array-field.tsx
```diff
@@ -222,8 +224,13 @@ export const ArrayField = (props: IGenericCustomFieldProps<IArrayFieldSchema>) =
 										type="button"
 										styleType="light"
 										danger
-										icon={removeButton?.icon ? renderIcon(removeButton.icon) : undefined}
 										onClick={() => handleRemoveSection(index)}
+										{...(removeButton?.icon
+											? {
+													as: ButtonWithIcon.Small,
```
weili-govtech, line 230: isnt `RemoveButton` already a `ButtonWithIcon.Small`?

PR #487, opened 2025-11-11
Handle non-numeric fileSize in uploadResponse

weili-govtech:
**Changes**
- Handle non-numeric fileSize in uploadResponse
- Added test case to cover scenario

**Additional information**
- For upload response without `fileUrl` / `dataURL`, component displays the file size based on the `fileSize` entry in `uploadResponse` 
- If fileSize is not a number, component is showing `NaN undefined`

### DIFF
--- a/src/components/fields/file-upload/file-upload-manager.ts
+++ b/src/components/fields/file-upload/file-upload-manager.ts
@@ -230,14 +230,13 @@ const FileUploadManager = (props: IProps) => {
 					mime: fileToInject.uploadResponse?.["mimeType"],
 					ext: fileToInject.uploadResponse?.["ext"],
 			  });
-		const { errorMessage: filesizeErrorMessage } = validateFileSize(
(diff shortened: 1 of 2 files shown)


src/components/fields/file-upload/file-upload-manager.ts
```diff
@@ -234,6 +234,10 @@ const FileUploadManager = (props: IProps) => {
 			rawFile?.size || fileToInject.uploadResponse?.["fileSize"]
 		);
 
+		let size = rawFile?.size || fileToInject.uploadResponse?.["fileSize"] || 0;
+		if (isNaN(size)) {
+			size = 0;
+		}
```
qroll: since there's also some validation on the file size being done, the sanitisation can be done first and used as the param for `validateFileSize` and file data

something like `"1024" > 1` will evaluate to true and return the size error

PR #498, opened 2025-11-20
[MOL-19835][TIEN] track user interaction using isInteracted instead o…

tientranngoc-alt:
**Change log**

**Issue**
- `isDirty` does not behave correctly: when `useCurrentDate` is enabled and the user manually clears the field and blurs, field takes the value as current date instead of undefined, and `isDirty` not change.
- This also prevents required validation from triggering and the error message "_This field is required_" does not appear.
- Can refer to this ticket [MOL-19835](https://sgtechstack.atlassian.net/jira/software/c/projects/MOL/boards/1861?selectedIssue=MOL-19835&sprints=27533)

**Change**
- Apply a local state `isInteracted` to track whether the user has interacted with the field.
- This replaces reliance on `isDirty` and ensures that once the user clears the input, the field set value `undefined`, allowing required validation


tientranngoc-alt:
I realized that `isDirty` only works when `defaultValues` are provided. The date field I used didn’t include any `defaultValues`, and after adding them, everything behaved correctly. 
So I’ll close this PR.

PR #499, opened 2025-11-26
[MOL-19808] Error when using conditional validation with conditional rendering

dangdev3003:
**Background**: Currently, when using conditional validation (when) with a field that requires conditional rendering, the app may crash because the validation tries to evaluate rules on a field that is not yet rendered. (Details in MOL-19808)
**Solution**: Add a safety check to prevent the crash.

### DIFF
--- a/src/context-providers/yup/helper.ts
+++ b/src/context-providers/yup/helper.ts
@@ -85,13 +85,9 @@ export namespace YupHelper {
 					when: fieldValidationConfig.when ? { ...fieldValidationConfig.when } : undefined,
 				};
 				Object.keys(parsedConfig.when).forEach((whenFieldId) => {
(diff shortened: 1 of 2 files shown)


dangdev3003:
@qroll  can you take a look at this MR?

src/context-providers/yup/helper.ts
```diff
@@ -85,6 +85,15 @@ export namespace YupHelper {
 					when: fieldValidationConfig.when ? { ...fieldValidationConfig.when } : undefined,
 				};
 				Object.keys(parsedConfig.when).forEach((whenFieldId) => {
+					// Check field existence
```
qroll: unnecessary comment

qroll (approved):

PR #510, opened 2025-12-09
CCUBE-2031: Enhance DS to support an image and component within the popover

hhung2710:
**Changes**
-   Enhance DS to support an image and component within the popover

**Additional information**

-   You may refer to this [CCUBE-2031](https://sgtechstack.atlassian.net/browse/CCUBE-2031)

### DIFF
--- a/src/__tests__/components/elements/popover/popover.spec.tsx
+++ b/src/__tests__/components/elements/popover/popover.spec.tsx
@@ -83,24 +83,11 @@ describe(UI_TYPE, () => {
 			hint: {
 				content: {
 					text: { uiType: "text-body", children: "Schema hint content" },
-				} as any,
+				},
 			},
 		});
 
 		fireEvent.click(screen.getByTestId(COMPONENT_TEST_ID));
 		expect(screen.getByText("Schema hint content")).toBeInTheDocument();
-	});
-
-	it("should handle enhanced content types without error", () => {
-		const enhancedContent = {
-			"text-element": { uiType: "text" as const, children: "Enhanced text" },
-			"span-wrapper": { uiType: "span" as const, children: "Enhanced span" },
-		};
-
-		expect(() => {
-			renderComponent({ hint: { content: enhancedContent as any } });
-		}).not.toThrow();
-
-		expect(screen.getByTestId(COMPONENT_TEST_ID)).toBeInTheDocument();
 	});
 });
--- a/src/stories/4-elements/popover/popover.stories.tsx
+++ b/src/stories/4-elements/popover/popover.stories.tsx
@@ -129,27 +129,6 @@ Default.args = {
 	hint: { content: "Hint" },
 };
 
-export const TooltipCustomisationWithSchema = Template("popover-component").bind({});
-TooltipCustomisationWithSchema.args = {
-	uiType: "popover",
-	children: "More info",
-	hint: {
-		content: {
-			text: {
-				children: [
-					"This",
-					"<mark>is</mark>",
-					"<u>a</u>",
-					"<strong>styled</strong>",
-					"<i>sentence</i>",
-					`<img src='https://assets.life.gov.sg/lifesg/logo-lifesg.svg' alt="logo">`,
-				],
-				uiType: "body-bl",
-			},
-		},
-	},
-};
-
 export const WithIcon = Template("popover-with-icon").bind({});
 WithIcon.args = {
 	uiType: "popover",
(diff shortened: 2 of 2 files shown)


src/components/elements/popover/types.ts
```diff
@@ -1,9 +1,17 @@
...
 
+export enum PopoverHintType {
+	DEFAULT = "default",
+	IMAGE = "image",
+	COMPONENT = "component",
+}
+
 interface IPopoverHint extends Pick<PopoverInlineProps, "position" | "customOffset" | "zIndex"> {
-	content: string;
+	content: string | ReactElement;
```
qroll: we should aim to keep elements serializable. using ReactElement is possible for custom apps with access to source code but not for JSON-only config

there are two options:
1. update `Sanitize` usage in popover to support img tags (you can refer to `src/components/elements/text/text.tsx`)
2. allow `content` to accept schema object for more control (you can refer to `topSection` and `bottomSection` of IReviewBaseSchema in `src/components/custom/review/types.ts`)

option 2 is more flexible and will likely be more suitable for the requestor's usecase since they seem to need a more complex layout

qroll: while we're here, can we check if `Sanitize` preserves the `alt` attribute? this is likely important for accessibility

src/components/elements/popover/popover.tsx
```diff
@@ -30,14 +30,25 @@ export const Popover = (props: IGenericElementProps<IPopoverSchema>) => {
 		return <Element />;
 	};
 
+	const renderPopoverContent = () => {
+		switch (hintContentType) {
+			case PopoverHintType.IMAGE:
+				return <img src={hintContent as string} alt="popover content" />;
```
qroll: looking at the proposed usage, for accessibility, the alt needs to be a accurate description for functional images. which means it needs to be configurable via a prop

check if `Sanitize` can support `alt` attributes (does not remove it from the final html), if not this may need to be exposed as a standalone element?

hhung2710: As i check the `Sanitize` does not remove the `alt` attributes from the final html so no the to exposed as a standalone element

src/components/elements/popover/popover.tsx
```diff
@@ -29,6 +31,21 @@ export const Popover = (props: IGenericElementProps<IPopoverSchema>) => {
 		const Element = Icons[icon];
 		return <Element />;
 	};
+	const sanitizeOptions: IOptions = {
+		allowedTags: sanitizeHtml.defaults.allowedTags.concat(["img"]),
+		allowedAttributes: false,
+	};
```
qroll: nitpick: can declare `sanitizeOptions` within `renderPopoverContent` since it's not referenced anywhere else

src/components/elements/popover/types.ts
```diff
@@ -1,9 +1,11 @@
 import { PopoverInlineProps } from "@lifesg/react-design-system/popover-v2";
 import * as Icons from "@lifesg/react-icons";
+import { ITextSchema, ITypographySchema } from "../text";
 import { IBaseElementSchema } from "../types";
+import { IInlineWrapperSchema } from "../wrapper";
 
 interface IPopoverHint extends Pick<PopoverInlineProps, "position" | "customOffset" | "zIndex"> {
-	content: string;
+	content: string | Record<string, ITextSchema | ITypographySchema | IInlineWrapperSchema>;
```
qroll: we can follow `src/components/custom/filter/types.ts` to support TBlockElementSchema and the default TWrapperSchema for rendering of other FEE elements

src/stories/4-elements/popover/popover.stories.tsx
```diff
@@ -129,6 +129,29 @@ Default.args = {
...
+export const PopoverImage = Template("popover-image").bind({});
+PopoverImage.args = {
+	uiType: "popover",
+	children: "More info",
+	hint: {
+		content: `<img src='https://assets.life.gov.sg/lifesg/logo-lifesg.svg' alt="logo">`,
+	},
+};
+
+export const PopoverComponent = Template("popover-component").bind({});
```
qroll: move together with TooltipCustomisation story

```suggestion
export const TooltipCustomisationWithSchema = Template("popover-component").bind({});
```

src/__tests__/components/elements/popover/popover.spec.tsx
```diff
@@ -77,4 +77,30 @@ describe(UI_TYPE, () => {
 
 		expect(label.compareDocumentPosition(icon)).toBe(Node.DOCUMENT_POSITION_FOLLOWING);
 	});
+
+	it("should render FEE schema content in hint", () => {
+		renderComponent({
+			hint: {
+				content: {
+					text: { uiType: "text-body", children: "Schema hint content" },
+				} as any,
```
qroll: no need for `as any`?

PR #519, opened 2025-12-29
Update pipeline script to exit on 401

qroll:
**Changes**

- job runs infinitely when error is encountered, updated the script to exit when the curl fails on 401
- remove unintended sleep when job status is complete
- see https://github.com/LifeSG/react-design-system/pull/920
-   [delete] branch


chongwyb (approved):
approve update of gitlab pipeline check

PR #535, opened 2026-02-02
[MOL-20239][DN]ESignatureField - Update handleChange func

dangnh64:
**Changes**
- ESignatureField: Update `handleChange` function to handle case that user clear the signature and save without drawing a new one.

**Additional information**
https://sgtechstack.atlassian.net/browse/MOL-20239

### DIFF
--- a/src/__tests__/common/helper.tsx
+++ b/src/__tests__/common/helper.tsx
@@ -118,7 +118,3 @@ export const FrontendEngineWithEventListener = (props: IFrontendEngineWithEventL
 
 	return <FrontendEngine {...otherProps} ref={formRef} />;
 };
-
-export const getButtonByName = (name: string): HTMLElement => {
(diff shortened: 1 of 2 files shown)


src/__tests__/common/data.ts
```diff
@@ -4,3 +4,6 @@ export const RESET_BUTTON_ID = "reset";
 export const RESET_BUTTON_LABEL = "Reset";
 export const ERROR_MESSAGE = "test error message";
 export const FRONTEND_ENGINE_ID = "test";
+export const EDIT_SIGNATURE_BUTTON_LABEL = "Edit signature";
+export const CLEAR_SIGNATURE_BUTTON_LABEL = "Clear";
+export const SAVE_SIGNATURE_BUTTON_LABEL = "Save";
```
qroll: this is specific to the esignature field, it shouldn't be in the common file

PR #538, opened 2026-02-09
Resolve MOL-19636

dangnh64:
**Changes**
- Add CallbacksProvider
- Add useCallbacks hook
- Update tab.tsx element

**Additional information**
https://sgtechstack.atlassian.net/browse/MOL-19636


qroll (changes requested):
actually FEE has an existing event handling system

you can look at `useFieldEvent` and its usage in `src/components/fields/button/button.tsx`

PR #539, opened 2026-02-16
[CCUBE-2080][XY] Expose customLabels for FileUpload

ninjasheeps:
**Changes**
- Expose `customLabels` prop for `uploadButtonLabel`
- Update dependencies to latest DS version to use `customLabels` from FileUpload

-   delete branch

### DIFF
--- a/src/components/fields/file-upload/file-upload.tsx
+++ b/src/components/fields/file-upload/file-upload.tsx
@@ -1,9 +1,8 @@
-import { FileUpload as DSFileUpload, FileItemProps } from "@lifesg/react-design-system/file-upload";
+import { FileItemProps, FileUpload as DSFileUpload } from "@lifesg/react-design-system/file-upload";
 import xor from "lodash/xor";
 import { Suspense, lazy, useCallback, useContext, useEffect, useRef, useState } from "react";
 import { useFormContext } from "react-hook-form";
 import * as Yup from "yup";
-import { IGenericFieldProps } from "..";
 import { FileHelper } from "../../../utils";
 import { useFieldEvent, useValidationConfig } from "../../../utils/hooks";
 import { IYupValidationRule } from "../../frontend-engine";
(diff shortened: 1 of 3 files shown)


src/components/fields/file-upload/file-upload.tsx
```diff
@@ -22,7 +21,7 @@ import {
 // lazy load to fix next.js SSR errors
 const FileUploadManager = lazy(() => import("./file-upload-manager"));
 
-export const FileUploadInner = (props: IGenericFieldProps<IFileUploadSchema>) => {
+export const FileUploadInner = (props: IFileUploadProps) => {
```
qroll: instead of introducing a new `IFileUploadSchema` type, could we add on the new schema props to `IFileUploadSchema` instead?

in fact we could extend the Pick directly
```
Pick<FileUploadProps, ... | "customLabels">
```

src/components/fields/file-upload/file-upload.tsx
```diff
@@ -43,6 +42,7 @@ export const FileUploadInner = (props: IGenericFieldProps<IFileUploadSchema>) =>
 			...otherSchema
 		},
 		warning,
+		customLabels,
```
qroll: from above, the customLabels would be destructured from `schema` instead

qroll (changes requested):
unit tests are failing, likely due to some breaking data-testid changes after bumping to the latest design system version. will need to update them accordingly!

src/__tests__/components/fields/range-select/range-select.spec.tsx
```diff
@@ -81,7 +81,7 @@ const ComponentWithSetSchemaButton = (props: { onClick: (data: IFrontendEngineDa
 };
 
 const getComponent = (): HTMLElement => {
-	return getField("button", "Select Select");
+	return screen.getByTestId("field-base");
```
qroll: this is the same as `getRangeSelector` 👀  I think we can replace everything to use `getRangeSelector`

PR #572, opened 2026-03-31
Checkbox group accessibility

benjaminLeongSK:
**Changes**
Added accessibility feature for checkbox default/toggle variant 
Cherry pick changes from [BL/543](https://github.com/LifeSG/web-frontend-engine/issues/543)

-   [delete] branch

### DIFF
--- a/src/components/fields/checkbox-group/checkbox-group.tsx
+++ b/src/components/fields/checkbox-group/checkbox-group.tsx
@@ -122,6 +122,7 @@ export const CheckboxGroup = (props: IGenericFieldProps<TCheckboxGroupSchema>) =
 							disabled={disabled ?? option.disabled}
 							name={checkboxId}
 							value={option.value}
+							focusableWhenDisabled={disabled}
 							checked={isCheckboxChecked(option.value)}
 							onChange={() => handleChange(option.value)}
 						/>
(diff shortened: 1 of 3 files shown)


src/components/fields/checkbox-group/checkbox-group.tsx
```diff
@@ -181,7 +183,9 @@ export const CheckboxGroup = (props: IGenericFieldProps<TCheckboxGroupSchema>) =
 	return (
 		<>
 			<Form.CustomField id={id} label={formattedLabel} errorMessage={error?.message}>
-				{customOptions?.styleType === "toggle" ? renderToggles() : renderCheckboxes()}
+				<div role="radiogroup" tabIndex={0}>
```
qroll: in the case of checkbox the wai aria patterns recommend using `group`. radiogroup is reserved for lists with only 1 selectable option

```suggestion
				<div role="group" tabIndex={0}>
```

qroll (changes requested):

src/components/fields/checkbox-group/checkbox-group.tsx
benjaminLeongSK: have to remove the tabIndex 
Screenshot 2026-04-25 at 9 17 22 PM

qroll (approved):

PR #576, opened 2026-04-10
[MOL-20560][NL] Add new event listeners-ready to location-field

namlh023:
**Changes**

In the new OS 2.0 enhancement, we need to display the location map immediately on page load in the mobile app. Normally, show-location-modal is triggered when the user clicks the location field, but now it runs on mount.

To support this, the location-field must register its show-location-modal event listener before the event is dispatched. This MR introduces a location-field listener-ready signal so consumers know when it’s safe to dispatch the event.

-   [delete] branch

**Changelog entry**

- Add new event: listeners-ready to location-field

**Additional information**

-   You may refer to this [MOL-20560](https://sgtechstack.atlassian.net/browse/MOL-20560)

### DIFF
--- a/src/components/fields/location-field/types.ts
+++ b/src/components/fields/location-field/types.ts
@@ -96,7 +96,7 @@ export type TSetCurrentLocationDetail = TLocationFieldDetail<ILocationCoord>;
 export type TLocationFieldErrorDetail = TLocationFieldDetail<TErrorType>;
 
 export type TLocationFieldEvents = {
-	"listeners-ready": CustomEvent;
+	"show-location-modal-ready": CustomEvent;
 	"get-current-location": CustomEvent;
 	"set-current-location": CustomEvent<TSetCurrentLocationDetail>;
 	error: CustomEvent<TLocationFieldErrorDetail>;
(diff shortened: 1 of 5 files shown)


weili-govtech:
shall we call it `mounted` event instead? `listeners-ready` imply all the listeners are ready but it is not the case

namlh023:
‘Mounted’ is a reasonable name, but it might suggest that it’s safe to dispatch the event as soon as the component mounts. In our case, there’s a short gap where the component is mounted but the listeners may not yet be fully registered (e.g., the show-location-modal listener is added in useEffect). The intent is to signal when the listeners are actually ready, so consumers know it’s safe to dispatch the event at that point.

In case if we choose ‘mounted’, then we can add a comment to clarify this behavior.

weili-govtech:
you have a point too, since we want to ensure `show-location-modal` event is ready, how about naming `show-location-modal-ready`?

src/components/fields/location-field/types.ts
```diff
@@ -120,7 +120,7 @@ export class GeolocationPositionErrorWrapper extends Error {
 /** fired after the location field registers its internal event listeners and is ready for external triggers */
```
weili-govtech:
```
/** fired after the location field registers the `show-location-modal`  event listener and is ready for external triggers */
```

weili-govtech (approved):

PR #577, opened 2026-04-10
[MOL-18749][MA] Add optional tooltip to image-upload component

mauw-paloit:
**Changes**

Added a `tooltip` prop to the `image-upload` field that renders a clickable element between the label and description.

-   [delete] branch

**Changelog entry**

-  Added optional `tooltip` prop to the `image-upload` field

**Additional information**

-   You may refer to this [ticket](https://sgtechstack.atlassian.net/browse/MOL-18749)

### DIFF
--- a/src/components/fields/image-upload/image-input/image-input.styles.ts
+++ b/src/components/fields/image-upload/image-input/image-input.styles.ts
@@ -34,9 +34,10 @@ export const TooltipWrapper = styled.button`
 	border: none;
 	padding: 0;
 	cursor: pointer;
+	color: ${Colour["text-primary"]};
 
 	&:hover {
-		opacity: 0.8;
+		color: ${Colour["text-hover"]};
 	}
 `;
 
@@ -46,7 +47,7 @@ export const TooltipIcon = styled.span`
 	justify-content: center;
 	width: 1rem;
 	height: 1rem;
-	color: ${Colour["text-primary"]};
+	color: inherit;
 
 	svg {
 		width: 100%;
@@ -55,7 +56,7 @@ export const TooltipIcon = styled.span`
 `;
 
 export const TooltipLabel = styled(Typography.BodyMD)`
-	color: ${Colour["text-primary"]};
+	color: inherit;
 `;
 
 export const Content = styled.div`
--- a/src/components/fields/image-upload/image-input/image-input.styles.ts
+++ b/src/components/fields/image-upload/image-input/image-input.styles.ts
@@ -24,6 +24,40 @@ export const Subtitle = styled(Typography.BodyBL)<SubtitleProps>`
 	margin-bottom: ${(props) => (props.$hasDescription ? Spacing["spacing-8"] : Spacing["spacing-16"])};
 `;
 
+export const TooltipWrapper = styled.button`
+	display: inline-flex;
+	align-items: center;
+	gap: ${Spacing["spacing-4"]};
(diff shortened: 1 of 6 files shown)


src/components/fields/image-upload/types.ts
```diff
@@ -33,6 +34,7 @@ export interface IImageUploadSchema<V = undefined>
 	capture?: TFileCapture | undefined;
 	multiple?: boolean | undefined;
 	imageReviewModalStyles?: string | undefined;
+	tooltip?: { label?: string | undefined; icon?: keyof typeof Icons | undefined; onClick: () => void } | undefined;
```
weili-govtech: schema needs to be JSON-compatible, hence it cannot support an `onClick` function
you can support it through field events instead (`dispatchFieldEvent`)

src/components/fields/image-upload/image-input/image-input.styles.ts
```diff
@@ -24,6 +24,40 @@ export const Subtitle = styled(Typography.BodyBL)<SubtitleProps>`
...
+	margin-top: ${Spacing["spacing-8"]};
+	margin-bottom: ${Spacing["spacing-16"]};
+	background: transparent;
+	border: none;
+	padding: 0;
+	cursor: pointer;
+
+	&:hover {
+		opacity: 0.8;
+	}
```
weili-govtech, line 41: need to confirm with ux if this is the intended hover effect, DS typically apply `text-hover` colour instead

src/components/fields/image-upload/image-input/image-input.styles.ts
```diff
@@ -24,6 +24,40 @@ export const Subtitle = styled(Typography.BodyBL)<SubtitleProps>`
...
+`;
+
+export const TooltipIcon = styled.span`
+	display: inline-flex;
+	align-items: center;
+	justify-content: center;
```
weili-govtech: redundant as the parent is already using flexbox

weili-govtech (changes requested):

src/components/fields/image-upload/types.ts
mauw-paloit: Thanks, I totally forgot about this. Updated to use dispatchFieldEvent

src/components/fields/image-upload/image-input/image-input.styles.ts
mauw-paloit: Checked with UX, updated to use text-hover

src/components/fields/image-upload/image-input/image-input.styles.ts
mauw-paloit: I kinda need this though for align-items and justify-content else the TooltipIcon element would adopt the default inline display

weili-govtech: but the svg within has 100% width and height, it is filling up all available space and has nothing to align to

src/components/fields/image-upload/types.ts
weili-govtech: use `click-tooltip` instead

mauw-paloit: updated to `click-tooltip`

src/components/fields/image-upload/image-input/image-input.styles.ts
mauw-paloit: fair point. Updated

weili-govtech (approved):

PR #588, opened 2026-04-22
[MOL-20562][NL] Support custom HTML markers

namlh023:
**Changes**

Custom HTML markers can be rendered by passing a markerHtml property for each marker. This HTML string is used to display a custom marker, enabling dynamic rendering based on pin data—for example, showing different icons or badges for various location types

-   delete branch

**Changelog entry**

- Support custom HTML markers via markerHtml for dynamic marker rendering
- Add a new event: click-selectable-pin

**Additional information**

-   You may refer to this [MOL-20562](https://sgtechstack.atlassian.net/browse/MOL-20562)


namlh023:
Support custom HTML markers via markerHtml instead of getMarkerHtml

src/stories/3-fields/location-field/location-field.stories.tsx
```diff
@@ -395,46 +491,30 @@ locationSelectionMode.args = {
...
-					pins: res.map(
-						(r) =>
-							({
-								lat: r.latitude,
-								lng: r.longitude,
-								resultListItemText: r.carParkName,
-								address: `${r.carParkName} (${r.carParkNumber})`,
-							} as IMapPin)
-					),
+				formRef.current?.dispatchFieldEvent("location-field", "set-selectable-pins", id, {
```
namlh023, line 522: zoom and center on the default address only after custom icons have been set (listen to set-selectable-pins event)

src/components/fields/location-field/location-modal/location-picker/helper.ts
```diff
@@ -19,6 +19,17 @@ export const markerFrom = (
 	});
 };
 
+export const markerFromHtml = ({ lat, lng }: IMapPin, html: string, className?: string | undefined): L.Marker => {
+	return L.marker([lat, lng], {
+		icon: L.divIcon({
+			html,
+			className: className || "",
+			iconSize: [1, 1],
```
mauw-paloit: this is an abnormally small iconSize and might cause issues with click hitboxes. Can refer to the `markerFrom` function to reference how it manages iconSize. Same for iconAnchor.

src/__tests__/components/fields/location-field/location-field.spec.tsx
```diff
@@ -33,6 +34,8 @@ import {
 	mockReverseGeoCodeResponse,
 	mockStaticMapDataUri,
 } from "./mock-values";
+import * as markerHelper from "../../../../components/fields/location-field/location-modal/location-picker/helper";
```
mauw-paloit: nit but could you sort the imports

src/components/fields/location-field/location-modal/location-picker/helper.ts
namlh023: Remove iconSize and let the consumer control the size of the custom marker. Set iconAnchor to (0, 0) by default. If we need to control the anchor position in the future, we can enhance it by adding width and height props as options.

PR #594, opened 2026-05-05
Upgrade to node 22

benjaminLeongSK:
**Changes**
Upgrading of node version
-   [delete] branch

**Changelog entry**

- Fix audit issues
- Bump dependencies 
- Replace deprecated test functions

**Additional information**

-   You may refer to this [CCUBE-1804](https://sgtechstack.atlassian.net/browse/CCUBE-1804)
-  For vulnerabilities in Dependabot alerts will have to wait till this branch is merged in for it to be updated and tackled


qroll (dismissed):

rollup.config.js
```diff
@@ -4,14 +4,15 @@ import json from "@rollup/plugin-json";
...
-import { terser } from "rollup-plugin-terser";
+import terser from "@rollup/plugin-terser";
 import typescript from "rollup-plugin-typescript2";
 
 const plugins = [
 	peerDepsExternal(), // Add the externals for me. [react, react-dom]
 	nodeResolve({ browser: true }),
 	commonjs(), // converts CommonJS to ES6 modules
 	typescript({
+		include: ["**/*.ts", "**/*.tsx", "**/*.cts", "**/*.mts"],
```
qroll, line 15: checking why we need to include the extensions explicitly

scripts/patch-jsdom.js
qroll, line 1: could we use patch-package to patch the library instead of using a custom script?

src/__tests__/components/fields/file-upload/file-upload.spec.tsx
```diff
@@ -83,19 +84,23 @@ interface IRenderAndPerformActionsOptions {
...
 		files = [],
+		getFileType,
 		inputType = "input",
 		uploadType = "base64",
 		headers = {},
 		onClick,
 	} = options;
+	jest.spyOn(FileHelper, "getType").mockImplementation(
+		getFileType || (() => Promise.resolve({ ext: "jpg", mime: "image/jpeg" }))
+	);
```
qroll, line 101: nit: let's keep the spies close together

add a new line after the options destructuring

package-lock.json
qroll, line 1: lockfile is out of sync

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -314,7 +329,7 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
 
 	const handlePencilErasable = (e: any) => {
 		const path = e.path as Path;
-		path.erasable = true; // Important for @erase2d/fabric
+		(path as ErasableFabricObject).erasable = true; // Important for @erase2d/fabric
```
qroll: hmm do you know why `erasable` was removed as a typing? is this the recommended method to restore the functionality?

qroll: I saw this in the library https://github.com/ShaMan123/erase2d/blob/d08a6cbcd9e7c9da5978395a83290dd65159b544/packages/fabric/README.md?plain=1#L29

qroll (changes requested):

rollup.config.js
benjaminLeongSK: without the explicit include list, rollup-plugin-typescript2 does not transform src/index.ts correctly and Rollup fails on the `export type` syntax.

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
benjaminLeongSK: hmm not sure why but its not part of Fabric’s core FabricObject / FabricImage typings anymore. the eraser behavior now comes from @erase2d/fabric, and that package adds erasable as an extra property at runtime. so locally there will be the type error. 

adding `"types": ["node_modules/@erase2d/fabric/types"]` to tsconfig doesnt work as well cos @erase2d/fabric@1.2.1 does not contain that types folder

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -191,13 +212,6 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
 				);
 			}
 
-			fabricCanvas.current.clipPath = new Rect({
-				left: img.left,
-				top: img.top,
-				width: img.getScaledWidth(),
-				height: img.getScaledHeight(),
-			});
-
```
benjaminLeongSK, line 200: We no longer need a Fabric clipPath because export now crops to the background image bounds using `getBoundingRect()`.

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -24,9 +27,19 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
 	const eraserBrush = useRef<EraserBrush>();
 	const gestures = useRef({
 		pinchStartAmount: 0,
+		pan: new Point(0, 0),
```
benjaminLeongSK, line 28: Keep our pan position separated so we can use Fabric's public `absolutePan` API instead of mutating `viewportTransform` directly

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -24,9 +27,19 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
...
 
+	const syncCanvasDimensions = () => {
+		if (!wrapperRef.current || !fabricCanvas.current) return;
+
+		fabricCanvas.current.setDimensions({
+			width: wrapperRef.current.clientWidth,
+			height: wrapperRef.current.clientHeight,
+		});
+	};
+
```
benjaminLeongSK, line 40: to keep the Fabric canvas dimensions in sync with the wrapper element before fitting or exporting the image.

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -51,15 +64,16 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
 	// recursively compress image until it fits limit
 	const toDataURLWithLimit = (limit = maxSizeInKb, quality = 1): string => {
 		if (fabricBackground.current) {
+			const backgroundBounds = fabricBackground.current.getBoundingRect();
```
benjaminLeongSK, line 65: using `getBoundingRect` bcos Fabric 7 position/scale calculations are more reliable through object bounds than raw `top`, `left`, or scaled width values.

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -146,15 +157,28 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
 	const resetZoomAndPosition = () => {
 		if (fabricCanvas.current) {
 			fabricCanvas.current.setZoom(1);
-
-			const vpt = fabricCanvas.current.viewportTransform || [0, 0, 0, 0, 0, 0];
-			vpt[4] = 0;
-			vpt[5] = 0;
-
+			gestures.current.pan = new Point(0, 0);
+			fabricCanvas.current.absolutePan(new Point(0, 0));
```
benjaminLeongSK, line 158: Reset both zoom and pan state when a new image is loaded so the image starts from its default fitted position.

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -146,15 +157,28 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
...
+		return new Point(Math.min(Math.max(pan.x, 0), maxPanX), Math.min(Math.max(pan.y, 0), maxPanY));
+	};
+
+	const applyPan = (pan: Point, zoom?: number) => {
+		if (!fabricCanvas.current) return;
+
+		gestures.current.pan = clampPanToCanvas(pan, zoom);
+		fabricCanvas.current.absolutePan(gestures.current.pan);
+	};
+
```
benjaminLeongSK: `clampPanToCanvas` ensures that while zooming the image doesnt anyhow move out of the edges. this is mainly to replace `viewPortTransform` here :
Image

to ensure that mobile zoom is behaving correctly

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -344,8 +359,15 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
...
+						const oldZoom = fabricCanvas.current.getZoom();
 						const [ox, oy] = origin;
-						fabricCanvas.current.zoomToPoint(new Point(ox, oy), zoom);
+						const nextPan = new Point(
+							((gestures.current.pan.x + ox) / oldZoom) * zoom - ox,
+							((gestures.current.pan.y + oy) / oldZoom) * zoom - oy
+						);
+
+						fabricCanvas.current.setZoom(zoom);
+						applyPan(nextPan, zoom);
```
benjaminLeongSK, line 388: preserve the pinch focal point by converting the existing pan from the old zoom level to the new zoom level. and use `setZoom` with our own clamped pan calculation instead of `zoomToPoint`,  because `zoomToPoint` updates the viewport transform directly.

benjaminLeongSK:
the canvas changes avoid direct `viewportTransform` mutation and use Fabric's public APIs (`setDimensions`, `centerObject`, `setZoom`, `absolutePan`) instead. pan is tracked separately so zoom and drag can be clamped consistently

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
qroll: then where are the library types located? does it work if we point to them? otherwise we can introduce our own custom `d.ts` file to augment the typings like the library

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -284,16 +296,15 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
...
 				// prevent drawing when gesturing
 				// clear accidental drawings while gesturing
 				try {
 					(fabricCanvas.current.freeDrawingBrush as any).onMouseUp({ e: {} });
 				} catch (e) {}
 				const canvasObjects = fabricCanvas.current.getObjects();
-				if (canvasObjects.length > 1) {
-					fabricCanvas.current.remove(canvasObjects[canvasObjects.length - 1]);
-				}
+				canvasObjects.slice(objectCountBeforeGesture).forEach((obj) => fabricCanvas.current?.remove(obj));
```
qroll: what does the `canvasObjects.slice` do here?

benjaminLeongSK: this removes any object accidentally created while gesturing by recording the object count before the gesture starts and then slice(objectCountBeforeGesture) gets only the newly added objects after that point.

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -110,8 +119,7 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
 				const oldBackgroundLeft = fabricBackground.current.left || 0;
 				const oldBackgroundTop = fabricBackground.current.top || 0;
 
-				fabricCanvas.current.setWidth(canvasWidth);
-				fabricCanvas.current.setHeight(canvasHeight);
+				syncCanvasDimensions();
 
 				const scale = Math.min(
 					canvasWidth / fabricBackground.current.getScaledWidth(),
```
weili-govtech: since `fabricBackground.current.getBoundingRect()` is more reliable now, shouldn't we use this instead of  `getScaledWidth()`

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
weili-govtech: reset pinch start amount too

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
qroll: slice doesn't mutate the array, so this line have any effect?

do we actually need to update `canvasObjects`?

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
weili-govtech: just tested, it does not work well, I am able to drag the photo edges beyond the screen boundaries

Image

package.json
```diff
@@ -79,7 +80,12 @@
 		"@rollup/plugin-json": "^4.1.0",
 		"@rollup/plugin-node-resolve": "^14.1.0",
 		"@storybook/addon-a11y": "^8.2.8",
-		"@storybook/addon-essentials": "^8.2.8",
+		"@storybook/addon-actions": "^8.6.18",
+		"@storybook/addon-backgrounds": "^8.6.14",
+		"@storybook/addon-controls": "^8.6.14",
+		"@storybook/addon-docs": "^8.6.14",
+		"@storybook/addon-toolbars": "^8.6.14",
+		"@storybook/addon-viewport": "^8.6.14",
```
weili-govtech:
```suggestion
		"@storybook/addon-backgrounds": "^8.6.18",
		"@storybook/addon-controls": "^8.6.18",
		"@storybook/addon-docs": "^8.6.18",
		"@storybook/addon-toolbars": "^8.6.18",
		"@storybook/addon-viewport": "^8.6.18",
```

weili-govtech (dismissed):

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
benjaminLeongSK: oh yeah you are right on this. shall revert back to the previous implementation

qroll (dismissed):

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
```diff
@@ -20,13 +20,24 @@ export const ImageEditor = forwardRef((props: IImageEditorProps, ref: ForwardedR
 	const canvasRef = useRef<HTMLCanvasElement>(null);
 	const fabricCanvas = useRef<FabricCanvas>();
 	const fabricBackground = useRef<FabricImage>();
+	const fittedImageBounds = useRef<{ left: number; top: number; width: number; height: number }>();
```
qroll, line 23: qn: where is `fittedImageBounds` assigned a value?

.storybook/main.ts
```diff
@@ -4,8 +4,12 @@ import path from "path";
 const config: StorybookConfig = {
 	stories: ["../src/stories/**/*.stories.@(ts|tsx)", "../src/stories/**/*.mdx"],
 	addons: [
-		"@storybook/addon-essentials",
-		"@storybook/addon-a11y",
+		"@storybook/addon-actions",
```
qroll, line 8: also restore the `"@storybook/addon-a11y"` plugin

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
benjaminLeongSK: my apologies it seems to be missing. it should have been assigned a value in `fitImageToCanvas`

qroll (approved):

src/components/fields/image-upload/image-review/image-editor/image-editor.tsx
qroll: reverted as this helps to clip any strokes that go out of bound

PR #595, opened 2026-05-09
CCUBE-1916

benjaminLeongSK:
**Changes**
Bump styled component and fix test warnings
-   [delete] branch

**Additional information**
- fix test warnings for styled component 

-   You may refer to this [CCUBE-1916](https://sgtechstack.atlassian.net/browse/CCUBE-1916)

### DIFF
--- a/src/components/elements/wrapper/field-wrapper.tsx
+++ b/src/components/elements/wrapper/field-wrapper.tsx
@@ -27,6 +27,17 @@ interface IProps {
 	warning?: string | undefined;
 }
 
+type TForwardedFieldStateProp = keyof Pick<ControllerFieldState, "isDirty" | "isTouched" | "invalid" | "isValidating">;
+
+const FIELD_STATE_PROPS_BY_UI_TYPE: Partial<Record<TFrontendEngineFieldSchema["uiType"], TForwardedFieldStateProp[]>> =
+	{
+		"contact-field": ["isDirty"],
+		"date-field": ["isDirty"],
+		"date-range-field": ["isDirty"],
+		"file-upload": ["isDirty", "isTouched"],
+		"image-upload": ["isDirty", "isTouched"],
+	};
+
 export const FieldWrapper = ({ Field, id, schema, warning }: IProps) => {
 	// =========================================================================
 	// CONST, STATE, REFS
@@ -129,10 +140,19 @@ export const FieldWrapper = ({ Field, id, schema, warning }: IProps) => {
 			value: getField(id),
 			warning,
 		};
-		return <Field schema={schema} {...fieldProps} {...fieldState} />;
+		return <Field schema={schema} {...fieldProps} {...getFieldStateProps(schema, fieldState)} />;
 	};
 
 	return <Controller control={control} name={id} shouldUnregister={true} render={renderField} />;
+};
+
+const getFieldStateProps = (schema: TFrontendEngineFieldSchema, fieldState: ControllerFieldState) => {
+	const forwardedProps = FIELD_STATE_PROPS_BY_UI_TYPE[schema.uiType] ?? [];
+
+	return {
+		error: fieldState.error,
+		...Object.fromEntries(forwardedProps.map((prop) => [prop, fieldState[prop]])),
+	};
 };
 
 const StyledSublabel = styled(Sanitize)`
--- a/src/components/elements/wrapper/field-wrapper.tsx
+++ b/src/components/elements/wrapper/field-wrapper.tsx
@@ -27,17 +27,6 @@ interface IProps {
 	warning?: string | undefined;
 }
 
-type TForwardedFieldStateProp = keyof Pick<ControllerFieldState, "isDirty" | "isTouched" | "invalid" | "isValidating">;
-
-const FIELD_STATE_PROPS_BY_UI_TYPE: Partial<Record<TFrontendEngineFieldSchema["uiType"], TForwardedFieldStateProp[]>> =
-	{
-		"contact-field": ["isDirty"],
-		"date-field": ["isDirty"],
-		"date-range-field": ["isDirty"],
-		"file-upload": ["isDirty", "isTouched"],
-		"image-upload": ["isDirty", "isTouched"],
-	};
-
 export const FieldWrapper = ({ Field, id, schema, warning }: IProps) => {
 	// =========================================================================
 	// CONST, STATE, REFS
@@ -140,19 +129,10 @@ export const FieldWrapper = ({ Field, id, schema, warning }: IProps) => {
 			value: getField(id),
 			warning,
 		};
-		return <Field schema={schema} {...fieldProps} {...getFieldStateProps(schema, fieldState)} />;
+		return <Field schema={schema} {...fieldProps} {...fieldState} />;
 	};
 
 	return <Controller control={control} name={id} shouldUnregister={true} render={renderField} />;
-};
-
-const getFieldStateProps = (schema: TFrontendEngineFieldSchema, fieldState: ControllerFieldState) => {
-	const forwardedProps = FIELD_STATE_PROPS_BY_UI_TYPE[schema.uiType] ?? [];
-
-	return {
-		error: fieldState.error,
-		...Object.fromEntries(forwardedProps.map((prop) => [prop, fieldState[prop]])),
-	};
 };
 
 const StyledSublabel = styled(Sanitize)`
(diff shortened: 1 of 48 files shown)


src/components/frontend-engine/frontend-engine.tsx
```diff
@@ -238,7 +249,9 @@ const FrontendEngineInner = forwardRef<IFrontendEngineRef, IFrontendEngineProps>
 export const FrontendEngine = forwardRef<IFrontendEngineRef, IFrontendEngineProps>((props, ref) => {
 	return (
 		<ContextProviders recaptchaSiteKey={props.recaptchaSiteKey}>
-			<FrontendEngineInner {...props} ref={ref} />
+			<StyleSheetManager shouldForwardProp={shouldForwardProp}>
```
qroll: this won't be a long term fix as we are moving away from styled components

can we find out specifically which components the errors are originating from and target the fixes there?

benjaminLeongSK: Screenshot 2026-05-11 at 4 12 06 PM

i see, we technically can target the fixes. its mainly caused by passing in extra props into styled button components (eg submit button). we can get around it by omitting out those props and it clears the warnings. what do you think

qroll: a lot more than these are being passed, they likely come from react-hook-form as `otherProps` in addition to the common schema options in `otherSchema`

Screenshot 2026-05-13 at 8 59 49 AM

I think there's no choice but to properly omit these props

- for `otherSchema`, because most of it is meant to be forwarded to the DS, it should be blacklist-based - we could have a common util to filter out the common schema options
- for `otherProps`, can help to evaluate if we could destructure specific properties or if a common util is also needed to whitelist/blacklist?

src/components/frontend-engine/prop-helper.ts
qroll, line 1: can this be moved to the utility folder instead?

src/components/frontend-engine/prop-helper.ts
```diff
@@ -0,0 +1,22 @@
+import omit from "lodash/omit";
+
+const COMMON_SCHEMA_PROP_KEYS = [
+	"children",
```
qroll: why `children`?

src/components/frontend-engine/prop-helper.ts
```diff
@@ -0,0 +1,22 @@
...
+	"clearBehavior",
```
qroll: `clearBehavior` looks to be specific to filter, should be destructured there directly

src/components/fields/select/select.tsx
```diff
@@ -6,23 +6,17 @@ import * as Yup from "yup";
...
-		id,
-		onChange,
-		schema: { label: _label, options, validation, ...otherSchema },
-		value,
-		warning,
-		...otherProps
-	} = props;
+	const { error, formattedLabel, id, onChange, schema, value, warning, ...otherProps } = props;
+	const { options, validation } = schema;
+	const selectProps = filterSchemaProps(schema, ["options"]);
```
qroll: drawback is that we are specifying the keys twice

I would propose to split into something like

```
const { commonSchema: { validation }, customSchema: { options, ...otherSchema } } = filterSchemaProps(schema);
```

qroll (changes requested):

src/components/frontend-engine/prop-helper.ts
benjaminLeongSK: hmm i added it in since its a schema-level prop.. that said i think it could be removed as well

qroll:
noting down the warnings that originate from the DS:
- `noBorder` warning for OTP verification
- unmounted state update for ElementWithDropdown

src/components/elements/wrapper/field-wrapper.tsx
qroll, line 1: if we ever need to access another prop in a field, we would to update this mapping which is not very efficient. and it is also not obvious from the typings that the props have been omitted.

similar to the schema, the component itself should ensure that it destructures the react-hook-props and only extracts the props that are used (similar to what was done for `text-field`

src/components/fields/chips/chips.tsx
```diff
@@ -16,16 +16,11 @@ export const Chips = (props: IGenericFieldProps<IChipsSchema>) => {
 	// =============================================================================
 	// CONST, STATE, REFS
 	// =============================================================================
+	const { error, formattedLabel, id, onChange, schema, value, warning, ...otherProps } = props;
```
qroll: I see that `otherProps` is being passed into `Form.CustomField` but it doesn't seem like any of those are used by the component on the DS. should be alright to remove it

src/components/fields/contact-field/contact-field.tsx
```diff
@@ -190,7 +183,7 @@ export const ContactField = (props: IGenericFieldProps<IContactFieldSchema>) =>
 	return (
 		<>
 			<Form.PhoneNumberInput
-				{...otherSchema}
+				{...inputProps}
 				{...otherProps}
```
qroll: should extract the required props from `otherProps` here as well. typically the FEE field should invoke react-hook-form's onChange (or onSelect etc.) and sometimes onBlur (if supported by the DS component). can check for all other fields

src/components/fields/histogram-slider/histogram-slider.tsx
```diff
@@ -25,20 +25,24 @@ export const HistogramSlider = (props: IGenericFieldProps<IHistogramSliderSchema
...
-			const min = Math.min(...bins.map((bin) => bin.minValue));
-			const max = Math.max(...bins.map((bin) => bin.minValue)) + interval;
-
-			resetField(id, { defaultValue: { from: min, to: max }, keepDirty: true });
-			setStateValue([min, max]);
+			resetField(id, { defaultValue: { from: rangeMin, to: rangeMax }, keepDirty: true });
+			setStateValue([rangeMin, rangeMax]);
 		} else {
-			setStateValue([value.from, value.to]);
+			setStateValue(getDisplayValue(value));
```
qroll, line 45: hmm can explain this change here?

src/components/fields/location-field/location-modal/location-search/location-search.tsx
```diff
@@ -128,9 +129,16 @@ export const LocationSearch = ({
 	// =============================================================================
 	// EFFECTS
 	// =============================================================================
+	useEffect(() => {
+		return () => {
+			isMounted.current = false;
+			reverseGeocodeAborter.current?.abort();
```
qroll, line 135: can explain why these changes are needed?

src/components/fields/text-field/text-field.tsx
```diff
@@ -224,14 +227,15 @@ export const TextField = (props: IGenericFieldProps<ITextFieldSchema | IEmailFie
 	return (
 		<>
 			<InputElement
-				{...otherSchema}
-				{...otherProps}
+				{...inputProps}
 				{...derivedAttributes}
 				id={id}
+				name={name}
```
qroll: we probably don't need to specify `name` since we are using javascript to handle the form 🤔

qroll (changes requested):

src/components/fields/histogram-slider/histogram-slider.tsx
benjaminLeongSK: This was to handle partially populated values. Previously, if `value` existed but `from` or `to` was undefined, we could pass `[undefined, undefined]` causing this warning to show up

src/components/fields/location-field/location-modal/location-search/location-search.tsx
benjaminLeongSK: this is to clean up in-flight reverse geocode work when the location search unmounts. `displayResultsFromLatLng` can start an async reverse geocode request, and if the modal closes before it resolves, the callback may still try to update state. which will trigger this warning as well. 

Screenshot 2026-05-14 at 10 29 06 PM

src/components/elements/wrapper/field-wrapper.tsx
```diff
@@ -129,7 +129,9 @@ export const FieldWrapper = ({ Field, id, schema, warning }: IProps) => {
 			value: getField(id),
 			warning,
 		};
-		return <Field schema={schema} {...fieldProps} {...fieldState} />;
+		const fieldStateProps = "referenceKey" in schema ? { error: fieldState.error } : fieldState;
```
qroll: hmm why only `error` prop for `referenceKey`? some custom components also need the field state. let's revert this

qroll (changes requested):

qroll (dismissed):

qroll (approved):

PR #596, opened 2026-05-11
[MOL-20692][AT] introduce recaptcha ready props to indicate the readiness of the component if the recaptcha has been set

thyAshley:
introduces a reCAPTCHA readiness signal so consumers can safely determine when reCAPTCHA is available before making protected API requests.

**Changes**

- Added a new isRecaptchaReady boolean to the reCAPTCHA context contract
- Computed readiness as:
  - true when no reCAPTCHA site key is configured (support existing user without any impact)
  - true when reCAPTCHA is configured and fully loaded
  - false otherwise
- Updated location search flow to guard API-dependent effects/actions until isRecaptchaReady is true.
-   delete branch

**Changelog entry**

-   Added `isRecaptchaReady` prop to reCAPTCHA context, allowing components to check reCAPTCHA availability before making protected API calls.

**Additional information**

-  Some location-related requests can execute before reCAPTCHA is actually usable causing an error modal to show for the user before they even use the component.
- This causes avoidable failures or race-condition behavior when token generation is attempted too early.
- The new readiness flag makes this explicit and prevents premature API calls.

### DIFF
--- a/src/components/fields/location-field/location-modal/location-search/location-search.tsx
+++ b/src/components/fields/location-field/location-modal/location-search/location-search.tsx
@@ -82,7 +82,7 @@ export const LocationSearch = ({
 	// CONST, STATE, REFS
 	// =============================================================================
 	const { addFieldEventListener, removeFieldEventListener, dispatchFieldEvent } = useFieldEvent();
-	const { getToken } = useRecaptcha();
+	const { isRecaptchaReady, getToken } = useRecaptcha();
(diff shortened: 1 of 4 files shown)


weili-govtech:
- what about errors like invalid recaptcha key? can we show the "Map not available" error modal
- add tests to guard this behaviour

thyAshley:
there is no validation of recaptcha key, if the key is wrong, we will still show map error modal as the api call will still fail.

weili-govtech (approved):

PR #610, opened 2026-06-02
[CCUBE-2100][IT] Add Linaria support in Storybook and Jest

iant-gds:
**Changes**

- [delete] branch
- Updated storybook `main.ts` to ensure storybook is able to process linaria css
- Created sample story and test file to ensure storybook and jest can process components styled using linaria

**Changelog entry**

- Added linaria libraries
- Added custom babel plugin


.storybook/main.ts
```diff
@@ -11,7 +11,22 @@ const config: StorybookConfig = {
 	features: { interactions: false, sidebarOnboardingChecklist: false },
 	staticDirs: ["../public"],
 	webpackFinal: async (config) => {
+		config.module?.rules?.unshift({
+			test: /\.(js|jsx|ts|tsx)$/,
+			include: [path.resolve(__dirname, "../src"), path.resolve(__dirname, "../stories")],
```
qroll: wrong path for `stories`. in this repo it is under `src` folder as well. we could omit the second entry

babel.config.json
```diff
@@ -1,6 +1,7 @@
 {
 	"presets": ["@babel/preset-env", ["@babel/preset-react", { "runtime": "automatic" }], "@babel/preset-typescript"],
```
qroll: also include the `@wyw-in-js/babel-preset`

qroll (changes requested):

babel.config.json
qroll: ugh sorry I was referencing the old config where we were still using the library's preset as part of the default set up. this can be removed since it has been superseded by our custom babel plugin

qroll (approved):

PR #612, opened 2026-06-04
[CCUBE-2101][IT] Add Linaria support in Rollup

iant-gds:
**Changes**
Description of changes...

- [delete] branch
- Updated `rollup.config.js` to handle linaria css

**Changelog entry**

- Updated rollup version to v4


rollup.config.js
```diff
@@ -20,10 +25,25 @@ const plugins = [
...
+		minimize: true,
+		sourceMap: true,
+	}),
 	image(),
 	json(),
 	terser(), // Helps remove comments, whitespace or logging codes
+	generatePackageJson({
+		outputFolder: "dist",
+		baseContents: (pkg) => ({
+			...pkg,
```
qroll: explicitly include the fields instead

things like `scripts`, `devDependencies` and `lint-staged` should not be part of the build for consumers

qroll (changes requested):

qroll:
`index.css` needs to be imported by the library's main entry point

package.json
```diff
@@ -122,10 +120,13 @@
...
 		"postcss-import": "^15.0.0",
 		"postcss-loader": "^7.3.3",
+		"postcss-url": "^10.1.4",
 		"prettier": "^2.7.1",
 		"react": "^17.0.2",
 		"react-dom": "^17.0.2",
-		"rollup": "^2.79.1",
+		"rollup": "^4.34.9",
+		"rollup-plugin-generate-package-json": "^3.2.0",
+		"rollup-plugin-lib-style": "^2.5.0",
```
qroll: hmm I see that `lib-style` was installed but not actually used

rollup.config.js
```diff
@@ -20,10 +23,48 @@ const plugins = [
...
+		outputFolder: "dist",
+		baseContents: (pkg) => ({
+			name: pkg.name,
+			version: pkg.version,
+			description: pkg.description,
+			typings: "./index.d.ts",
+			main: "./cjs/index.js",
+			module: "./index.js",
+			style: "./index.css",
+			sideEffects: ["*.css", "./index.js", "./cjs/index.js"],
```
qroll: I'm not familiar with sideEffects - wondering why `index.js` also needs to be listed here?

rollup.config.js
```diff
@@ -20,10 +23,48 @@ const plugins = [
...
+			if (chunk.isEntry) {
+				const cssImport = options.format === "cjs" ? 'require("./index.css");' : 'import "./index.css";';
+				return cssImport + "\n" + code;
+			}
+			return null;
+		},
+	},
+	generatePackageJson({
+		outputFolder: "dist",
+		baseContents: (pkg) => ({
```
qroll, line 55: can we also include the other informational fields from package.json? in particular, keywords/homepage/bugs/license/repository/engines

I am not 100% sure if the `postinstall` script needs to be included as well - could we check if it was it intended to run in the consumer project?

qroll (changes requested):

rollup.config.js
iant-gds: Have removed it, css styling solved itself after fixing the imports

rollup.config.js
iant-gds: Based on existing implementations, i don't see `postinstall` script being used

qroll (approved):

PR #613, opened 2026-06-05
[CCUBE-2136][GZ] setup e2e test

ghazwan-gt:
### Changes
- Setup playwright to run E2E test
- Provided story examples for both component and frontend engine
- Provided simple test examples for both component and frontend engine


e2e/nextjs-app/app/globals.css
```diff
@@ -0,0 +1,18 @@
+@import url("/cdn/react-design-system/v3/css/main.css");
+@import url("/cdn/react-design-system/v3/css/atkinson-hyperlegible-next.css");
```
qroll: we can skip the set up for other themes and use lifesg as the default

e2e/tests/utils/fixtures.ts
```diff
@@ -0,0 +1,17 @@
+import { test as base } from "@playwright/test";
+import { StoryPage, type TStoryPageOptions } from "./story-page";
+
+export const forComponent = StoryPage.forComponent;
+
+export const test = base.extend<{
+	story: StoryPage;
+	storyOptions: TStoryPageOptions;
+}>({
+	storyOptions: [{ component: "elements/divider", story: "default" }, { option: true }],
```
qroll: hmm could we not apply a default setting for the `storyOptions`?

scripts/e2e-setup.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -euo pipefail
+
+PROJECT_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+
+pushd "$PROJECT_DIR" >/dev/null
+
+echo "[E2E Setup] Installing root dependencies"
+npm ci
```
qroll: split `npm ci` to a standalone step (so that this script can run locally without reinstalling node modules)

scripts/e2e-setup.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -euo pipefail
+
+PROJECT_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+
+pushd "$PROJECT_DIR" >/dev/null
```
qroll: can explain what this step does?

.github/workflows/e2e-playwright.yml
```diff
@@ -0,0 +1,41 @@
...
+          node-version: 22
+          cache: npm
+
+      - name: Setup E2E Environment
+        run: ./scripts/e2e-setup.sh
+
+      - name: Run E2E Tests
+        env:
+          CI: true
+        run: npm run test-e2e:ci
```
qroll: hmm doesn't this end up running the `e2e-setup.sh` script twice?

.github/workflows/e2e-playwright.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          node-version: 22
+          cache: npm
+
+      - name: Setup E2E Environment
+        run: ./scripts/e2e-setup.sh
+
+      - name: Run E2E Tests
+        env:
+          CI: true
```
qroll: I was under the impression that CI is set to true by default in the runner so the env config is not needed

e2e/nextjs-app/app/globals.css
qroll, line 1: still need to include the base files:

```
@import url("/cdn/react-design-system/v3/css/main.css");
@import url("/cdn/react-design-system/v3/css/open-sans.css");
```

e2e/nextjs-app/app/globals.css
```diff
@@ -0,0 +1,8 @@
+:root {
+	color-scheme: light;
+}
```
qroll: is this needed?

e2e/nextjs-app/app/layout.tsx
```diff
@@ -0,0 +1,37 @@
+import type { Metadata } from "next";
+import type { ReactNode } from "react";
+import "./globals.css";
+
+const CDN_STYLE_PATHS = [
```
qroll: why explicit use `link` instead of `@import` in `globals.css`?

e2e/nextjs-app/app/layout.tsx
```diff
@@ -0,0 +1,37 @@
...
+	"/cdn/react-design-system/v3/css/main.css",
+	"/cdn/react-design-system/v3/css/atkinson-hyperlegible-next.css",
```
qroll: to exclude the font files outside of `open-sans.css`

e2e/nextjs-app/app/page.tsx
```diff
@@ -0,0 +1,8 @@
+export default function HomePage() {
+	return (
+		<section className="panel" data-testid="home-page">
```
qroll: this classname is not defined anywhere

introduce a `page.module.css` if custom styles are needed

e2e/nextjs-app/next.config.ts
```diff
@@ -0,0 +1,27 @@
...
+const isCI = process.env.CI === "true";
+
+const workspaceRoot = path.resolve(__dirname, "../..");
+const frontendEngineSourceRelative = "./src/index.ts";
+const designSystemRelative = "./e2e/nextjs-app/node_modules/@lifesg/react-design-system";
+
+const nextConfig = {
+	reactStrictMode: true,
+	experimental: {
+		externalDir: true,
```
qroll, line 38: what does this option do?

e2e/nextjs-app/tsconfig.json
qroll, line 1: does this mean we don't need to configure the `paths` to handle conflicting react versions?

e2e/tests/frontend-engine/example/example.e2e.spec.ts
qroll, line 1: I am thinking we won't need functional testing for validation - it's well covered by the unit tests

could this test file be reworked to a simple sanity check? we can render a basic schema in the story and assert that the form element can render in the browser successfully

playwright.config.ts
qroll, line 1: we should include the docker set up as well, to reduce flakiness with font rendering

playwright.config.ts
```diff
@@ -0,0 +1,29 @@
...
+		trace: "retain-on-failure",
+	},
+	projects: [
+		{
+			name: "chromium",
+			use: { ...devices["Desktop Chrome"] },
+		},
+	],
+	webServer: {
+		command: "npm --prefix e2e/nextjs-app run dev",
```
qroll: suggest to build the actual NextJS application so we can test against the "production" behaviour

qroll (changes requested):

e2e/nextjs-app/app/layout.tsx
ghazwan-gt: Last time I did `@import`, it doesn't go through the cdn proxy

e2e/nextjs-app/app/page.tsx
ghazwan-gt: Yup, this is a leftover from initial setup

e2e/nextjs-app/tsconfig.json
qroll: ah I recalled the other function of the custom path. it allows us to make changes to FEE and have the Typescript changes reflect immediately in the nextjs app for local development

e2e/nextjs-app/next.config.mjs
```diff
@@ -0,0 +1,52 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+// This config is authored as an ES module (`.mjs`), so Node does not provide
+// `__filename` and `__dirname` automatically. We derive them explicitly from
+// `import.meta.url` so all path calculations below remain stable regardless of
+// where the process is launched from.
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
```
qroll: I noticed that the DS uses `next.config.ts`  rather than `.mjs` and doesn't need to implement this

we can do the same for FEE

scripts/e2e-ci.sh
```diff
@@ -0,0 +1,43 @@
...
+PROJECT_DIR="$(cd "$(dirname "$0")/.." && pwd)"
+
+# Change into the repository root and silence pushd output to keep
+# logs focused on setup actions rather than directory stack noise.
+pushd "$PROJECT_DIR" >/dev/null
+
+# In CI, we build the library first so downstream E2E runs against the
+# built artifact, which mirrors release behavior. Locally we skip the
+# build to keep iteration fast while developing tests.
+if [ "${CI:-}" = "true" ]; then
```
qroll: since this script is meant for running in CI, we can skip the condition checks for this flag

scripts/e2e-ci.sh
```diff
@@ -0,0 +1,43 @@
...
+	echo "[E2E Setup] Building FEE package (CI mode)"
+	./scripts/build.sh
+else
+	echo "[E2E Setup] Skipping build (dev mode - using source)"
+fi
+
+# Install dependencies for the Next.js E2E app.
+echo "[E2E Setup] Installing Next app dependencies"
+npm --prefix e2e/nextjs-app ci
```
qroll: ah we can move this to the `Install dependencies` step in the github action as well

scripts/e2e-ci.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# In CI we explicitly install the freshly built dist output into the
+# Next.js app so tests validate the packaged artifact, not source files.
+if [ "${CI:-}" = "true" ]; then
+	echo "[E2E Setup] Installing built FEE package into Next app"
+	npm --prefix e2e/nextjs-app install --no-save "$PROJECT_DIR/dist"
```
qroll: install the generated tar file `dist/lifesg-web-frontend-engine-<version>.tgz` instead

package.json
```diff
@@ -8,11 +8,17 @@
 	"scripts": {
 		"build": "npm run rollup",
 		"build-storybook": "storybook build",
+		"build-check": "npm run rollup",
```
qroll: `build-check` script is not needed

package.json
```diff
@@ -8,11 +8,17 @@
...
+		"test-e2e": "npm run test-e2e:setup && playwright test --ui",
+        "test-e2e:docker": "docker compose up --build",
+        "test-e2e:setup": "./scripts/e2e-setup.sh",
+		"test-e2e:ci": "npm run test-e2e:setup && ./scripts/e2e-ci.sh",
+		"post:build": "node ./scripts/post-build.js",
```
qroll: `post:build` script is not needed

qroll (changes requested):

e2e/nextjs-app/app/layout.tsx
qroll: I see, FDS has the same problem when running in turbopack. I guess it's only working now because of iut normally runs in webpack mode

qroll (approved):

PR #633, opened 2026-07-01
[MISC][GZ] shared dirty state test suite and render helper

ghazwan-gt:
**Changes**
- Introduce `dirtyStateTestSuite` and `createRenderComponent` utils to reduce boilerplate across field integration tests
- Converts test files to utilize the utils, removing ~2000 lines of duplicated test code. Small amount of test cases are left untouched since those doesn't follow the same pattern.

**Checklist**
- [x] All test suites pass
- [x] TypeScript compilation passes

### DIFF
--- a/src/__tests__/common/tests/labels.ts
+++ b/src/__tests__/common/tests/labels.ts
@@ -1,6 +1,7 @@
 import { fireEvent, screen } from "@testing-library/react";
 
-export const labelTestSuite = (renderComponent: (overrideField: unknown) => void) =>
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const labelTestSuite = (renderComponent: (overrideField?: any) => void) =>
 	describe("labels", () => {
 		const getComputedStyle = window.getComputedStyle.bind(window);
 
--- a/src/__tests__/common/tests/labels.ts
+++ b/src/__tests__/common/tests/labels.ts
@@ -1,7 +1,7 @@
 import { fireEvent, screen } from "@testing-library/react";
+import { TRenderComponent } from "../types";
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const labelTestSuite = (renderComponent: (overrideField?: any) => void) =>
+export const labelTestSuite = (renderComponent: TRenderComponent) =>
 	describe("labels", () => {
 		const getComputedStyle = window.getComputedStyle.bind(window);
 
--- a/src/__tests__/components/custom/array-field/array-field.spec.tsx
+++ b/src/__tests__/components/custom/array-field/array-field.spec.tsx
@@ -17,8 +17,7 @@ import {
 	getSubmitButton,
 	getSubmitButtonProps,
 } from "../../../common";
-import { warningTestSuite } from "../../../common/tests/warnings";
-import { dirtyStateTestSuite } from "../../../common/tests";
+import { dirtyStateTestSuite, warningTestSuite } from "../../../common/tests";
 
 const SUBMIT_FN = jest.fn();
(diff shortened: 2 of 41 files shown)


src/__tests__/common/tests/labels.ts
```diff
@@ -1,6 +1,7 @@
 import { fireEvent, screen } from "@testing-library/react";
 
-export const labelTestSuite = (renderComponent: (overrideField: unknown) => void) =>
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const labelTestSuite = (renderComponent: (overrideField?: any) => void) =>
```
qroll: qn: why does the param need to be changed from `unknown` to `any`? `unknown` seems to work fine without type errors

src/__tests__/components/custom/array-field/array-field.spec.tsx
```diff
@@ -18,6 +18,7 @@ import {
 	getSubmitButtonProps,
 } from "../../../common";
 import { warningTestSuite } from "../../../common/tests/warnings";
+import { dirtyStateTestSuite } from "../../../common/tests";
```
qroll: could we include warnings in the barrel index and combine the imports?

src/__tests__/components/custom/iframe/iframe.spec.tsx
```diff
@@ -46,7 +47,16 @@ interface IRenderOptions {
 	eventListener?: ((this: Element, ev: Event) => any) | undefined;
 }
 
-const renderComponent = (
+const renderComponent = createRenderComponent<IIframeSchema>({
+	componentId: COMPONENT_ID,
+	baseSchema: {
+		referenceKey: UI_TYPE,
+		src: IFRAME_SRC,
+	},
```
qroll, line 55: hmm I don't see much savings here since we have to redeclare the schema, and it feels a little awkward that to reference the full schema we have to access it from `renderComponent.schema`. could we still retain the common JSON_SCHEMA in the spec file and pass that wholesale to `createRenderComponent`?

not sure if there are other patterns I missed!

src/__tests__/components/fields/otp-verification-field/otp-verification-field.spec.tsx
```diff
@@ -286,65 +286,11 @@ describe(UI_TYPE, () => {
...
-						onClick={handleClick}
-					/>
-				);
-				fireEvent.change(getPhoneNoInput(), { target: { value: "98765432" } });
-				fireEvent.click(getResetButton());
-				fireEvent.click(screen.getByRole("button", { name: "Custom Button" }));
-
-				expect(formIsDirty).toBe(false);
-			});
+		dirtyStateTestSuite({
```
qroll, line 288: I see that this was applied for phone but not email, why is that the case?

src/__tests__/common/tests/labels.ts
ghazwan-gt: Fixed the typing!

src/__tests__/components/custom/iframe/iframe.spec.tsx
ghazwan-gt: The helper function now return the render fn and schema, making it more conventional. It behaves pretty much similar to the original code, which clone the schema. The difference is now we reduce the amount of boilerplate lines to do this

src/__tests__/components/fields/otp-verification-field/otp-verification-field.spec.tsx
ghazwan-gt: I've included the email to also utilize the suite, alongside with other fields

qroll (approved):
nice!

PR #663, opened 2026-07-17
[CCUBE-2199][RYN] migrate e-signature component

ryan-nguyen-t:
**Type of changes**

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [ ] Breaking change (fix or feature that would cause existing apis or functionality to change)
-   [x] Tests (improvements to unit tests or E2E tests)
-   [ ] Other (technical improvements, refactoring, or changes that don't fall into the above categories)

**Description of changes**

-   Link to [ticket](https://sgtechstack.atlassian.net/browse/CCUBE-2199)

**Checklist**

-   [ ] Changes follow the project guidelines in [CONTRIBUTING.md](https://github.com/LifeSG/react-design-system/blob/master/CONTRIBUTING.md)
-   [ ] Updated documentation
-   [ ] Added/updated unit tests

**Screenshots**


src/components/fields/e-signature-field/e-signature-field.styles.ts
```diff
@@ -1,14 +1,13 @@
...
-export const ESignatureWrapper = styled.div`
+export const eSignatureWrapper = css`
 	&:not(:last-child) {
 		margin: 0 0 ${Spacing["spacing-32"]};
 	}
 `;
 
-export const ErrorWrapper = styled.div`
+export const errorWrapper = css`
 	margin: -${Spacing["spacing-16"]} 0 ${Spacing["spacing-32"]};
```
qroll: negative margin does not work with css variables

with reference to what was done in FDS, apply it as `calc(${...} * -1)` instead

e2e/tests/components/fields/e-signature-field/e-signature-field.e2e.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { expect, forComponent, test } from "../../../utils/fixtures";
+
+const withStory = forComponent("fields/e-signature-field");
+
+const drawSignature = async (page: Parameters<Parameters<typeof test>[1]>[0]["page"]) => {
+	const canvas = page.locator(".upper-canvas");
```
qroll: as with the previous PRs, declare a story object to put the locators in one place

e2e/tests/components/fields/e-signature-field/e-signature-field.e2e.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+	});
+
+	test.describe(() => {
+		test.use({
+			storyOptions: {
+				...withStory("with-upload"),
+			},
+		});
+
+		test("Upload error display", async ({ story }) => {
```
qroll, line 86: looking at the implementation of the error handling, you should include the case where `loadErrorCount >= 3`

qroll (changes requested):

qroll:
if the signature drawing behaviour is flaky, you probably need to wait for the modal to complete its animation before drawing

you can refer to the DS functional tests

e2e/tests/components/fields/e-signature-field/e-signature-field.e2e.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+			await story.goto();
+
+			await test.step("Verify initial load error appears", async () => {
+				await expect(story.locators.loadErrorMessage).toBeVisible();
+				await expect(story.locators.tryAgainButton).toBeVisible();
+			});
+
+			await test.step("Click try again to reach refresh alert threshold", async () => {
+				await story.locators.tryAgainButton.click();
+				await story.locators.tryAgainButton.click();
```
ghazwan-gt, line 123: Is it expected to have no wait or assertion between clicks?

src/components/fields/e-signature-field/e-signature-field.tsx
```diff
@@ -191,52 +193,60 @@ export const ESignatureField = (props: IGenericFieldProps<IESignatureFieldSchema
...
-			</ErrorWrapper>
+			</div>
 		);
 	};
 
 	return (
-		<ESignatureWrapper>
+		<div className={clsx(styles.eSignatureWrapper, className)}>
 			<Form.ESignature
 				{...otherSchema}
```
ghazwan-gt, line 250: `otherSchema` doesn't include className, should we pass it here as well?

qroll: ~~we should pass the className selector in only 1 place, and the parent element should generally be the target~~

qroll: sorry after checking DS, className is passed directly to the esignature component, not the form wrapper. so to keep with the original behaviour, in FEE the className should be passed to `Form.ESignature` as well

good catch!

src/components/fields/e-signature-field/e-signature-field.tsx
```diff
@@ -21,7 +22,7 @@ export const ESignatureField = (props: IGenericFieldProps<IESignatureFieldSchema
 		formattedLabel,
 		id,
 		onChange,
-		schema: { label: _label, uiType: _uiType, upload, validation, ...otherSchema },
+		schema: { label: _label, uiType: _uiType, upload, validation, className, ...otherSchema },
```
qroll: if the `className` prop is not being chained or use for other purposes, we can omit the destructuring from `otherSchema`

e2e/tests/components/fields/e-signature-field/e-signature-field.e2e.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+			tryAgainButton: page.getByRole("button", { name: "Please try again." }),
+			loadRefreshAlert: page.getByTestId("load-refresh-alert"),
+		};
+	}
+
+	public async drawSignature() {
+		await this.locators.canvas.waitFor({ state: "visible" });
+		await this.locators.signatureModal.evaluate((el) =>
+			Promise.all(el.getAnimations().map((animation) => animation.finished))
+		);
```
qroll: instead of checking the animation property, can check for the `stable` element state instead

copy the implementation in DS to a helper in AbstractStoryPage:

```
export const waitForAnimationEnd = async (locator: Locator) => {
    const handle = await locator.elementHandle();
    await handle?.waitForElementState("stable");
    await handle?.dispose();
};
```

e2e/tests/components/fields/e-signature-field/e-signature-field.e2e.spec.ts
```diff
@@ -0,0 +1,135 @@
...
+
+			await test.step("Click try again to reach refresh alert threshold", async () => {
+				await story.locators.tryAgainButton.click();
+				await expect(story.locators.tryAgainButton).toBeVisible();
+				await story.locators.tryAgainButton.click();
+			});
+
+			await test.step("Verify refresh alert appears after 3 failed load attempts", async () => {
+				await expect(story.locators.loadRefreshAlert).toBeVisible();
+				await story.page.evaluate(() => document.fonts.ready);
```
qroll: why do we need to check `document.fonts.ready`?

e2e/tests/components/fields/e-signature-field/e-signature-field.e2e.spec.ts
ghazwan-gt: The assertion seems weird. Does the try again button hidden after each click?

src/components/custom/review/review.tsx
qroll, line 1: exclude this change

e2e/tests/components/fields/e-signature-field/e-signature-field.e2e.spec.ts
qroll: the error display may change by the third click. so here i think it's ok to verify that the try button still remains

ghazwan-gt (approved):

qroll:
resolve the merge conflict

qroll (approved):

PR #708, opened 2026-08-04
[CCUBE-2263][GZ] change radio button layout column sizing

ghazwan-gt:
**Type of changes**

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [X] Breaking change (fix or feature that would cause existing apis or functionality to change)
-   [ ] Tests (improvements to unit tests or E2E tests)
-   [X] Other (technical improvements, refactoring, or changes that don't fall into the above categories)

**Description of changes**

-   Link to [ticket](Switch to t-shirt sizing for radio-button columns)
-   Change the radio button layout column sizing from `mobile, tablet, desktop` to `sm, lg, xl`

**Checklist**

-   [X] Changes follow the project guidelines in [CONTRIBUTING.md](https://github.com/LifeSG/react-design-system/blob/master/CONTRIBUTING.md)
-   [X] Updated documentation
-   [X] Added/updated unit tests

**Screenshots**


src/components/fields/radio-button/types.ts
```diff
@@ -21,14 +21,14 @@ export interface IImageButtonOption extends IRadioButtonOption {
 
 export type TRadioToggleLayoutType = "horizontal" | "vertical";
 
-export type TBreakpoint = "mobile" | "tablet" | "desktop";
+export type TBreakpoint = "sm" | "lg" | "xl";
```
qroll: should expand to `"xxs" | "xs" | "sm" | "md" | "lg" | "xl"`

but for e2e tests we can continue to test only a subset (we can assume logic is the same)

qroll (changes requested):

src/components/fields/radio-button/types.ts
ghazwan-gt: Ooh we aim to cover for all the available sizes. Alright, will update

src/stories/3-fields/radio-button/radio-toggle-button.stories.tsx
```diff
@@ -41,10 +41,10 @@ const meta: Meta = {
...
 			table: {
 				type: {
-					summary: `{styleType: "toggle", indicator?: boolean, border?: boolean, layoutType?: "horizontal" | "vertical", layoutColumns?: number | { mobile?: number, tablet?: number, desktop?: number }, minItemWidth?: number | { mobile?: number, tablet?: number, desktop?: number }, stretch?: boolean}`,
+					summary: `{styleType: "toggle", indicator?: boolean, border?: boolean, layoutType?: "horizontal" | "vertical", layoutColumns?: number | { sm?: number, lg?: number, xl?: number }, minItemWidth?: number | { sm?: number, lg?: number, xl?: number }, stretch?: boolean}`,
```
qroll: could the docs be updated as well to reflect the new keys?

src/components/fields/radio-button/radio-button.tsx
```diff
@@ -28,17 +28,20 @@ import {
...
 ): T => {
 	if (value === undefined || value === null) return defaultValue;
 	if (typeof value !== "object") return value as T;
-	const { mobile, tablet, desktop } = value as { mobile?: T; tablet?: T; desktop?: T };
-	if (breakpoint === "mobile") return mobile ?? tablet ?? desktop ?? defaultValue;
-	if (breakpoint === "tablet") return tablet ?? desktop ?? mobile ?? defaultValue;
-	return desktop ?? tablet ?? mobile ?? defaultValue;
+	const responsive = value as Partial<Record<TBreakpoint, T | undefined>>;
+	const idx = BREAKPOINT_ORDER.indexOf(breakpoint);
+	const searchOrder = [...BREAKPOINT_ORDER.slice(0, idx + 1).reverse(), ...BREAKPOINT_ORDER.slice(idx + 1)];
```
qroll: why does this still need to search through `...BREAKPOINT_ORDER.slice(idx + 1)` here?

ghazwan-gt: This is to handle when the current breakpoint has no match to the smaller breakpoints.
For example, current breakpoint is `lg`, but the provided options is `{xl: XXX}`, it will be resolved to `XXX`

qroll: hmm I would expect it to only inherit from the smaller breakpoints (similar to the col div logic in the DS)

so e.g. `md` would consider this order of precedence: `md` > `sm` > `xs` > `xxs`, if none, fallback to default
- for width, the default width
- for column count, auto-fit

ghazwan-gt: I misunderstand the previous logic. I thought it was going over to the larger breakpoint. Will do the DS approach instead

qroll: have clarified with the previous dev and you're right actually, the logic was meant to take the "nearest" from the parent

however we agreed that we should keep the behaviour consistent for generic use cases

qroll (approved):

PR #722, opened 2026-09-02
[DCUBESDQLR-2832][GR] add revalidation for when-dependent fields

Govind2112:
**Type of changes**

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [ ] Breaking change (fix or feature that would cause existing apis or functionality to change)
-   [x] Tests (improvements to unit tests or E2E tests)
-   [x] Other (technical improvements, refactoring, or changes that don't fall into the above categories)

**Description of changes**

-   Link to [DCUBESDQLR-2832](https://sgtechstack.atlassian.net/browse/DCUBESDQLR-2832)
-   Link to [Figma](url) : Not applicable
-   Fixed stale validation for Yup when conditions where dependent field errors were not updated when source fields changed.
- Added useWhenRevalidation hook to watch when dependencies and trigger dependent field re-validation.
- Added getWhenDependencies utility (recursive support for nested then / otherwise conditions).
- Integrated re-validation hook in FieldWrapper so behavior applies across Frontend Engine fields.
- Exported hook via index.ts.
- Added unit tests for dependency extraction (use-when-revalidation.spec.ts).

**Checklist**

-   [ ] Changes follow the project guidelines in [CONTRIBUTING.md](https://github.com/LifeSG/react-design-system/blob/master/CONTRIBUTING.md)
-   [ ] Updated documentation
-   [x] Added/updated unit tests

**Screenshots**

- No screenshots included because this is a validation behavior fix with no visual UI changes.

### DIFF
--- a/src/utils/hooks/use-when-revalidation.ts
+++ b/src/utils/hooks/use-when-revalidation.ts
@@ -1,0 +1,53 @@
+import { useEffect, useRef } from "react";
+import { useFormContext, useWatch } from "react-hook-form";
+import { IYupValidationRule } from "../../context-providers";
+
+export const getWhenDependencies = (validationRules: IYupValidationRule[] | undefined): string[] => {
+	if (!validationRules?.length) return [];
+
+	const deps: string[] = [];
+
+	const extractDeps = (rules: IYupValidationRule[]) => {
+		rules.forEach((rule) => {
+			if (rule.when && typeof rule.when === "object") {
+				Object.entries(rule.when).forEach(([key, condition]) => {
+					if (!deps.includes(key)) {
+						deps.push(key);
+					}
+
+					if (condition.then?.length) {
+						extractDeps(condition.then as IYupValidationRule[]);
+					}
+
+					if (condition.otherwise?.length) {
+						extractDeps(condition.otherwise as IYupValidationRule[]);
+					}
+				});
+			}
+		});
+	};
+
+	extractDeps(validationRules);
+	return deps;
+};
+
+export const useWhenRevalidation = (fieldId: string, validationRules: IYupValidationRule[] | undefined): void => {
+	const { trigger } = useFormContext();
+	const dependencies = getWhenDependencies(validationRules);
+
+	const watchedValues = useWatch({
+		name: dependencies,
+		disabled: dependencies.length === 0,
+	});
+
+	const isMountedRef = useRef(false);
+
+	useEffect(() => {
+		if (!isMountedRef.current) {
+			isMountedRef.current = true;
+			return;
+		}
+
+		void trigger(fieldId);
+	}, [watchedValues, fieldId, trigger]);
+};
--- a/src/__tests__/utils/hooks/use-when-revalidation.spec.ts
+++ b/src/__tests__/utils/hooks/use-when-revalidation.spec.ts
@@ -1,0 +1,48 @@
+import { getWhenDependencies } from "../../../utils/hooks/use-when-revalidation";
+import { IYupValidationRule } from "../../../context-providers";
+
+describe("getWhenDependencies", () => {
+	it("should return empty array when validation rules are undefined", () => {
+		expect(getWhenDependencies(undefined)).toEqual([]);
+	});
+
+	it("should return empty array when validation rules are empty", () => {
+		expect(getWhenDependencies([])).toEqual([]);
+	});
+
+	it("should return empty array when no when rules exist", () => {
+		expect(getWhenDependencies([{ required: true }])).toEqual([]);
+	});
+
+	it("should return single dependency from a when rule", () => {
+		const rules: IYupValidationRule[] = [
+			{
+				when: {
+					fieldA: { is: "yes", then: [{ required: true }] },
+				},
+			},
+		];
+		expect(getWhenDependencies(rules)).toEqual(["fieldA"]);
+	});
+
+	it("should return multiple dependencies from a single when rule", () => {
+		const rules: IYupValidationRule[] = [
+			{
+				when: {
+					fieldA: { is: "yes", then: [{ required: true }] },
+					fieldC: { is: "active", then: [{ required: true }] },
+				},
+			},
+		];
+		expect(getWhenDependencies(rules)).toEqual(["fieldA", "fieldC"]);
+	});
+
+	it("should not duplicate dependencies across multiple when rules", () => {
+		const rules: IYupValidationRule[] = [
+			{ when: { fieldA: { is: "yes", then: [{ required: true }] } } },
+			{ when: { fieldA: { is: "no", then: [{ required: false }] } } },
+			{ when: { fieldC: { is: "active", then: [{ required: true }] } } },
+		];
+		expect(getWhenDependencies(rules)).toEqual(["fieldA", "fieldC"]);
+	});
+});
--- a/src/components/elements/wrapper/field-wrapper.tsx
+++ b/src/components/elements/wrapper/field-wrapper.tsx
@@ -14,7 +14,13 @@ import {
 	useFormContext,
 } from "react-hook-form";
 import styled from "styled-components";
-import { useFormSchema, useFormValues, useIsomorphicDeepLayoutEffect, useValidationConfig } from "../../../utils/hooks";
+import {
+	useFormSchema,
+	useFormValues,
+	useIsomorphicDeepLayoutEffect,
+	useValidationConfig,
+	useWhenRevalidation,
+} from "../../../utils/hooks";
 import { IComplexLabel } from "../../fields";
 import { TFrontendEngineFieldSchema } from "../../frontend-engine/types";
 import { Sanitize } from "../../shared";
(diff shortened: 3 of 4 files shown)


src/utils/hooks/use-when-revalidation.ts
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		const { isSubmitted } = formState;
+
+		if (isSubmitted) {
+			if (revalidationMode === "onChange") {
+				void trigger(fieldId);
+			}
+		} else {
+			if (validationMode === "onChange" || validationMode === "onTouched") {
```
sevinlai-dev: validationMode normally points to the field itself. For example, field A has "onTouched" validationMode so field A validation is only triggered when field A is touched. But in this enhancement, field A might not be touched and its validation is already triggered because field B is touched. This validation behaviour might come as a surprise to existing application at production as it's not straightforward

I think the best solution is to somehow make use of react-hook-form own feature to trigger the validation though I'm not sure how can this be done.

Second best solution I can think of is to disable this type of validation by default and only enable it if specifically enabled in config. To make it simpler, just validate without checking validationMode. Though overall, it might complicate the feature but ensures no surprises at production

Govind2112: Updated the implementation to make when-dependent revalidation opt-in via enableWhenRevalidation. This keeps existing production behavior unchanged by default. When enabled, it reuses the existing RHF/Yup validation flow through trigger(fieldId) and only adds dependency-driven revalidation for when rules.

Govind2112: The conditional validation is implemented in a dedicated revalidation hook, not in the schema builder; it watches when dependencies and calls trigger(fieldId) to refresh validation without altering the base Yup schema generation. This is a safe enhancement, not a breaking change, as long as enableWhenRevalidation remains gated and the existing schema flow remains intact.

src/components/frontend-engine/types.ts
```diff
@@ -78,6 +78,8 @@ export interface IFrontendEngineData<V = undefined, C = undefined> {
 	revalidationMode?: TRevalidationMode | undefined;
 	/** Validation strategy before a user submits the form (onSubmit event) */
 	validationMode?: TValidationMode | undefined;
+	/** Enables seamless revalidation for fields with yup `when` dependencies when source fields change */
+	enableWhenRevalidation?: boolean | undefined;
```
weili-govtech: no need to control whether to enable this feature, we are fixing a bug, it should be always enabled

src/utils/hooks/use-when-revalidation.ts
weili-govtech, line 1: I am ok with extracting the logic to a hook, but how does this hook work with the existing logic in https://github.com/LifeSG/web-frontend-engine/blob/main/src/context-providers/yup/helper.ts#L203

do they interfere with each other? 
can we combine both logic?

do note that the existing logic merely constructs the Yup validation schema, it still defer to react-hook-form to control when validation takes place according to `validationMode` and `revalidationMode`

src/utils/hooks/use-when-revalidation.ts
```diff
@@ -0,0 +1,58 @@
+import { useEffect, useRef } from "react";
+import { useFormContext, useWatch } from "react-hook-form";
+import { IYupValidationRule } from "../../context-providers";
+import { useFormSchema } from "./use-form-schema";
+
+export const getWhenDependencies = (validationRules: IYupValidationRule[] | undefined): string[] => {
```
weili-govtech: if you are still going ahead with using a hook, you can explore merging  `getWhenDependencies()` into `useWhenRevalidation()` so the consumer just needs to call it once instead of passing this output into the hook

src/components/frontend-engine/types.ts
Govind2112: Agreed, The when dependency revalidation should be always enabled so the dependent field revalidates whenever the source field changes.

src/utils/hooks/use-when-revalidation.ts
Govind2112: They do not interfere, because they operate at different layers such as,

- YupHelper.buildSchema() creates the conditional schema (when / then / otherwise) and is the schema definition layer.
- useWhenRevalidation() is the React Hook Form runtime layer that decides when to force validation for the field after a dependency field changes.

To be confirmed:

- No, they are not redundant.
- No, they do not conflict.
- Yes, the dependency extraction could be shared in a utility, but the actual revalidation trigger should stay in the hook because it depends on RHF form state and trigger(fieldId)
- keep schema construction in YupHelper
- keep runtime revalidation in useWhenRevalidation
- optionally extract only the dependency-discovery logic into a shared helper if desired

src/utils/hooks/use-when-revalidation.ts
Govind2112: Yes, that is a reasonable cleanup if the goal is to simplify the call site.

Current pattern:
extract dependencies
pass them into the hook

Simplified pattern:
the hook computes the dependencies internally from the field validation rules

src/utils/hooks/use-when-revalidation.ts
weili-govtech: so why / why not combine them in a single place? what are your considerations?

I think it's ok to consult AI, but do digest their answers and think further

Govind2112: We should not fully merge them because schema definition and validation timing are different responsibilities, so keep Yup rule building in the schema helper and RHF trigger timing in the hook, sharing only dependency discovery if needed

Pros:
1. One central place for dependency logic.
2. Less duplicate code if dependency extraction is repeated.

Cons: 
1. Makes the code harder to reason about and test.
2. Create validation loops or over-triggering when dependency fields change.
3. Couples Yup schema construction to RHF form state and trigger logic.
4. Mixes two different responsibilities: schema definition and runtime validation timing.

weili-govtech: leaning in my thoughts, I am not sure if I understood correctly, I'll be happy to discuss further:

> Makes the code harder to reason about and test.

could you elaborate on that? is it because they are potentially two different things?
`YupHelper` is a pure function and more easily testable than a hook that requires a react component

> Create validation loops or over-triggering when dependency fields change.

`YupHelper` already has similar triggering logic [here](https://github.com/LifeSG/web-frontend-engine/blob/main/src/context-providers/yup/helper.ts#L225), are you referring to a risk in misuse of the validation rule? on the flip, wont `useWatch` + `useEffect` seem more likely to cause an over-trigger?

> Couples Yup schema construction to RHF form state and trigger logic.

`YupHelper` defines the validation behaviour as part of the schema construction, the trigger logic is part of the validation logic (where does RHF form state refers to btw?)

> Mixes two different responsibilities: schema definition and runtime validation timing.

similar to point 3, it is part of the validation behaviour

Govind2112: could you elaborate on that? is it because they are potentially two different things?
YupHelper is a pure function and more easily testable than a hook that requires a react component

`Yes,  that’s exactly the distinction YupHelper.buildSchema() is a pure schema-construction function, while a hook that watches RHF state and calls trigger() is a runtime orchestration layer, so combining them makes the code harder to reason about and test because one place would now be responsible for both rule definition and validation timing.`

Govind2112: YupHelper already has similar triggering logic [here](https://github.com/LifeSG/web-frontend-engine/blob/main/src/context-providers/yup/helper.ts#L225), are you referring to a risk in misuse of the validation rule? on the flip, wont useWatch + useEffect seem more likely to cause an over-trigger?

`The over-trigger risk is in the RHF runtime watcher layer, not in Yup’s declarative when schema rules.”`

Govind2112: YupHelper defines the validation behaviour as part of the schema construction, the trigger logic is part of the validation logic (where does RHF form state refers to btw?)

`The coupling risk is not from validating rules themselves, but from making a schema-definition utility depend on RHF runtime state and trigger behavior`

weili-govtech: > Yes,  that’s exactly the distinction YupHelper.buildSchema() is a pure schema-construction function, while a hook that watches RHF state and calls trigger() is a runtime orchestration layer, so combining them makes the code harder to reason about and test because one place would now be responsible for both rule definition and validation timing.

I believe you misunderstood the approach? we are not talking about a 1:1 merge - it wont be feasible to merge a hook directly into `YupHelper`, there is a need to change the logic to fit into a pure function.

regarding the validation timing, `YupHelper` runs when a validation takes place according to `validationMode` / `revalidationMode`, I dont think there is additional timing to account for?

> The over-trigger risk is in the RHF runtime watcher layer, not in Yup’s declarative when schema rules.”

indeed, `YupHelper` relies on the when schema rule, hence the over-trigger is lesser. in contrast, the hook relies on RHF's `watcher` layer + React's `useEffect`, isnt that more risky?

> The coupling risk is not from validating rules themselves, but from making a schema-definition utility depend on RHF runtime state and trigger behavior

similar to point 1, I am is not referring to using hooks in `YupHelper`.

Govind2112: Initially I also thought of keeping hooks in `YupHelper` and later, I understand your queries but based on the ticket requirement we don’t have a proper `revalidate` source field when dependent fields change under the existing `validationMode` or `revalidationMode`. 

So my understanding is that `YupHelper` should stay as pure schema construction, while dependency extraction remains a pure function and the revalidation logic is handled separately with a small useWatch plus trigger() hook.

Combining logic into one place is not good approach. I guess.

Govind2112: Conditional fields now re-validate automatically when the fields they depend on change, using the existing Yup when logic and a reverse dependency map.

weili-govtech: thanks I think the changes are appropriate, I would have preferred to keep everything in YupHelper to keep everything pure but this is an appropriate, readable alternative too

PR #735, opened 2026-09-17
Hardening pass: input validation, origin checks, and default-safe field behaviour — main (v3-alpha)

shengxi-gt:
# Hardening pass — main (v3-alpha)

## ⚠️ Breaking changes

Six changes alter default runtime behaviour unconditionally, no opt-out, for every existing schema as soon as this ships:

1. **`allowedFileOrigins` is now required for `FileUpload` prefill.** A prefilled `fileUrl` is no longer fetched unless `allowedFileOrigins` is explicitly set. Checked all in-scope consumer repos; none currently rely on the old default.
2. **Prefilled files without a verifiable source now default to rejected.** New `trustProvidedFileMetadata` flag, defaults `false`. Files with no `dataURL`/`fileUrl` are now flagged as unverified instead of silently accepted.
3. **`Iframe` `postMessage` origin check is now hard default-on.** Messages from a mismatched origin are silently dropped.
4. **500-character cap on schema-authored regex input.** Inputs longer than 500 chars are treated as non-matching regardless of the actual pattern result. Affects `matches`, `notMatches`, and `filenameMatches` rules.
5. **MaskedField `maxLength` now defaults to 500 whenever `maskRegex` is set** and no `max`/`length` rule already narrows it.
6. **Pattern strings not wrapped in `/pattern/flags` slashes are no longer silently skipped.** They previously landed in a catch block and validation was skipped entirely (value always passed). The shared parsing helper now falls back to treating the whole string as a raw pattern and applies it — any schema relying on the old silent no-op will now get real validation applied.

Narrower breaking changes (only affect a consumer that relied on the specific permissive behaviour being fixed): stricter `sanitize-html` configuration may now strip attributes that previously passed through; the ButtonField URL scheme check may block non-standard schemes; custom modal style strings now have `@import`/`url()` stripped.

## Changes covered

- `sanitize-html` configuration tightened in Typography, FilterCheckbox, Popover
- Iframe `postMessage` now validates origin by default
- Location-field search query now escaped before use in `RegExp`
- `FileUpload` prefill now requires an explicit origin allowlist
- Prefilled file metadata now requires explicit opt-in to trust
- OTP `state` field documented as client-asserted (no code change)
- Schema regex inputs now capped at 500 characters
- MaskedField `maxLength` defaults to 500 when `maskRegex` is set
- Custom modal style strings now strip `@import` and `url()` references
- ImageUpload `matches` rule no longer blocks valid submissions (pre-existing bug fix)
- Playwright e2e dev-server configuration hardened

## Test plan

- Full Jest suite + `eslint` + `tsc --noEmit` pass on this branch.


.github/workflows/trigger-gitlab-pipeline.yml
```diff
@@ -16,10 +16,15 @@ jobs:
...
+          HEAD_REF: ${{ github.head_ref }}
+          REF_NAME: ${{ github.ref_name }}
+          PR_TITLE: ${{ github.event.pull_request.title }}
+          HEAD_COMMIT_MSG: ${{ github.event.head_commit.message }}
         run: |
-          [[ $GITHUB_EVENT_NAME = "pull_request" ]] && BRANCH_NAME="${{ github.head_ref }}" || BRANCH_NAME="${{ github.ref_name }}"
+          [[ $GITHUB_EVENT_NAME = "pull_request" ]] && BRANCH_NAME="$HEAD_REF" || BRANCH_NAME="$REF_NAME"
 
-          [[ $GITHUB_EVENT_NAME = "pull_request" ]] && COMMIT_MSG="${{ github.event.pull_request.title }}" || COMMIT_MSG=$(echo -e "${{ github.event.head_commit.message }}" | head -n 1)
+          [[ $GITHUB_EVENT_NAME = "pull_request" ]] && COMMIT_MSG="$PR_TITLE" || COMMIT_MSG=$(echo -e "$HEAD_COMMIT_MSG" | head -n 1)
```
thanhpt0105-gt: https://github.com/LifeSG/react-icons/blob/main/.github/workflows/trigger-gitlab-pipeline.yml#L16-L22
Can follow this PR for the COMMIT_MSG

docker-compose.yml
```diff
@@ -1,6 +1,7 @@
 services:
     playwright-tests:
         build: .
-        network_mode: "host"
+        ports:
+            - "127.0.0.1:3010:3010"
```
thanhpt0105-gt, line 5: asguard-0004 should be false positive and don't need to update

thanhpt0105-gt (changes requested):

docker-compose.yml
shengxi-gt: Ran a quick, controlled reproduction to check this rather than just argue from theory:

Started the actual Playwright automation server bound to `0.0.0.0` (the pre-fix config) on my machine, then connected to it from a separate script using my machine's real LAN IP (not `127.0.0.1`) — no different from another device on the same WiFi doing it:

```
CONNECTED VIA LAN IP (192.168.1.9), NOT localhost. Remotely-read page title: PWNED-FROM-LAN-IP
```

It drove a real headless browser end-to-end (navigate + read page content) from that non-loopback connection — confirming this isn't just "a port is open," it's unauthenticated remote browser control.

Then re-ran with the fix (`--host 127.0.0.1`, same effect as `ports: ["127.0.0.1:3010:3010"]` in this PR):

```
BLOCKED as expected: browserType.connect: WebSocket error: connect ECONNREFUSED 192.168.1.9:3011
Local connection still works. Title: local-still-works
```

Same port, only the bind address changed — the LAN path is refused at the kernel level (not a firewall rule that could be misconfigured), and the legitimate local caller (the actual test runner, which always connects via `127.0.0.1`) is completely unaffected.

You're right that on GitHub-hosted CI runners this specific exposure window is low-risk — single-tenant, ephemeral VM, nothing else on that box to reach it during a run. But this same `docker-compose.yml` is also how this gets run locally and potentially on shared/self-hosted runners, and the fix has zero cost to the legitimate path. I'd lean toward keeping it as defense-in-depth given there's no downside — happy to revert if you feel strongly otherwise.

.github/workflows/trigger-gitlab-pipeline.yml
shengxi-gt: updated

src/components/custom/iframe/iframe.tsx
```diff
@@ -130,11 +137,17 @@ export const Iframe = (props: IGenericCustomFieldProps<IIframeSchema>) => {
 	// =========================================================================
 	// POSTMESSAGE HANDLERS
 	// =========================================================================
+	// only messages from the origin derived from `src` are accepted; a child iframe that
+	// navigates to a different origin before posting back will need to do so via the
+	// original src origin's window, or the message is dropped
```
wenheloh, line 142: reminder to add this potentially breaking change in changelog later, and maybe need TLDR like "if whatever you're serving using iframe and it navigated to another domain, it might break"

src/__tests__/components/fields/location-field/location-search-helper.spec.ts
wenheloh, line 1: normally we will put the test file at the same directory of the target that we're testing, why is this test file here?

qroll:
the sanitize-html bump is going to break jest and potentially downstream consumers 

see this comment for more info: https://github.com/LifeSG/web-frontend-engine/pull/721#issuecomment-5696689232

would suggest to leave it out of this ticket and tackle separately

src/components/custom/filter/filter-checkbox/filter-checkbox.tsx
```diff
@@ -89,7 +90,7 @@ export const FilterCheckbox = (props: IGenericCustomFieldProps<IFilterCheckboxSc
 			onSelect={handleChange}
 			valueExtractor={(item) => (isParentOption(item) ? item.key : item.value)}
 			labelExtractor={(item) => (
-				<Sanitize inline sanitizeOptions={{ allowedAttributes: false }}>
+				<Sanitize inline sanitizeOptions={{ allowedAttributes: sanitizeHtml.defaults.allowedAttributes }}>
```
qroll, line 93: if `sanitizeHtml.defaults.allowedAttributes` is the default, we can just omit `sanitizeOptions` here?

src/__tests__/components/fields/location-field/location-search-helper.spec.ts
qroll: FDS and FEE have a separate `__tests__` directory rather than co-locating test files 😬

src/__tests__/components/fields/location-field/location-search-helper.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+	});
+
+	it("should not hang or throw on a pathological regex-injection query", () => {
+		const maliciousQuery = "(a+)+$";
+
+		const start = Date.now();
+		expect(() =>
+			boldResultsWithQuery([buildResult("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa!")], maliciousQuery)
+		).not.toThrow();
+		expect(Date.now() - start).toBeLessThan(1000);
```
qroll, line 29: jest just seems to hang indefinitely on this test case, this time check doesn't seem useful

src/context-providers/yup/helper.ts
```diff
@@ -210,13 +188,23 @@ export namespace YupHelper {
...
-						try {
-							yupSchema = (yupSchema as Yup.StringSchema).matches(
-								new RegExp(matches[1], matches[2]),
-								rule.errorMessage
-							);
-						} catch (error) {
+						// "matches" tests the field's own value as a string. Applying it to a non-string
+						// schema (e.g. an array field) would run RegExp.test() against that value coerced
+						// to a string instead — near-guaranteed to fail and reject an otherwise-valid
+						// submission, not a meaningful validation of anything.
```
qroll, line 194: we can simplify the comment here

```suggestion
						// "matches" tests the field's own value as a string; skip for non-string values
```

src/utils/regex-helper.ts
```diff
@@ -0,0 +1,28 @@
...
+	/** upper bound on the string length tested against a schema-authored regex pattern.
+	 * A genuinely pathological pattern can still backtrack catastrophically below this length,
+	 * but bounding it caps the worst-case cost for the common cases (filenames, form field values,
+	 * short keystroke buffers) this is applied against. */
+	export const MAX_SAFE_PATTERN_INPUT_LENGTH = 500;
+
+	/** parses a `/pattern/flags`-style string into a RegExp, matching the convention already used
+	 * across the codebase for schema-authored regex config. Returns undefined instead of throwing
+	 * on an invalid pattern. */
+	export const parseMatchesPattern = (pattern: string): RegExp | undefined => {
```
qroll, line 11: can we align this to `validation-schema-generator`'s [implementation](https://github.com/LifeSG/validation-schema-generator/pull/106/changes#diff-82247497872d1591ec35076bd7c0aca7158a7f9e8e39626043e89d5654de698a)?

and rename to `compile`

src/utils/regex-helper.ts
```diff
@@ -0,0 +1,28 @@
+export namespace RegexHelper {
+	/** upper bound on the string length tested against a schema-authored regex pattern.
+	 * A genuinely pathological pattern can still backtrack catastrophically below this length,
+	 * but bounding it caps the worst-case cost for the common cases (filenames, form field values,
+	 * short keystroke buffers) this is applied against. */
+	export const MAX_SAFE_PATTERN_INPUT_LENGTH = 500;
```
qroll, line 6: align to `validation-schema-generator`, 1000 was used

src/context-providers/yup/helper.ts
```diff
@@ -210,13 +188,23 @@ export namespace YupHelper {
...
+							console.warn(`error applying "${ruleKey}" condition to ${yupSchema.type} schema`);
+							break;
+						}
+						const regex = RegexHelper.parseMatchesPattern(rule.matches);
+						if (regex) {
+							yupSchema = (yupSchema as Yup.StringSchema).test({
+								name: "matches",
+								message: rule.errorMessage,
+								test: (value) =>
+									value === undefined || value === null || RegexHelper.safeTestRegex(regex, value),
```
qroll, line 205: also return true for empty string case

src/components/fields/file-upload/types.ts
qroll, line 1: would argue against this change, the value are provided by the client/server in the first place

PR #739, opened 2026-09-18
Hardening pass: input validation, origin checks, and default-safe field behaviour — main (v3-alpha)

shengxi-gt:
# Hardening pass — main (v3-alpha)

## ⚠️ Breaking changes

Four changes alter default runtime behaviour unconditionally, no opt-out, for every existing schema as soon as this ships:

1. **\`Iframe\` \`postMessage\` origin check is now hard default-on.** Messages from a mismatched origin are silently dropped.
2. **1000-character cap on schema-authored regex input.** Inputs longer than 1000 chars are treated as non-matching regardless of the actual pattern result. Affects \`matches\`, \`notMatches\`, and \`filenameMatches\` rules.
3. **MaskedField \`maxLength\` now defaults to 500 whenever \`maskRegex\` is set** and no \`max\`/\`length\` rule already narrows it.
4. **Pattern strings not wrapped in \`/pattern/flags\` slashes are no longer silently skipped.** They previously landed in a catch block and validation was skipped entirely (value always passed). The shared parsing helper now falls back to treating the whole string as a raw pattern and applies it — any schema relying on the old silent no-op will now get real validation applied.

Narrower breaking changes (only affect a consumer that relied on the specific permissive behaviour being fixed): stricter \`sanitize-html\` configuration may now strip attributes that previously passed through; the ButtonField URL scheme check may block non-standard schemes; custom modal style strings now have \`@import\`/\`url()\` stripped.

## Changes covered

- \`sanitize-html\` configuration tightened in Typography, FilterCheckbox, Popover
- Iframe \`postMessage\` now validates origin by default
- Location-field search query now escaped before use in \`RegExp\`
- OTP \`state\` field documented as client-asserted (no code change)
- Schema regex inputs now capped at 1000 characters
- MaskedField \`maxLength\` defaults to 500 when \`maskRegex\` is set
- Custom modal style strings now strip \`@import\` and \`url()\` references
- ImageUpload \`matches\` rule no longer blocks valid submissions (pre-existing bug fix)
- Playwright e2e dev-server configuration hardened

## Test plan

- Full Jest suite + \`eslint\` + \`tsc --noEmit\` pass on this branch.


shengxi-gt:
> jest just seems to hang indefinitely on this test case, this time check doesn't seem useful

You were right that the old version hung — the query was passed directly to `new RegExp()` without escaping. In this PR we added `escapeRegExp(query)` before building the `RegExp` (line 12 of `helper.ts`), so `(a+)+$` becomes the literal pattern `\(a\+\)\+\$`. No backtracking; the test completes in microseconds. We also moved the spec file to the `location-modal/location-search/` subdirectory as part of the reorganisation.

shengxi-gt:
> would argue against this change, the value are provided by the client/server in the first place

The concern isn't distrust of the originating server — it's that between the server delivering the prefilled payload and the browser submitting the form, the data lives in the page where a user (or MITM proxy) can modify it. The frontend can verify files it fetches (`fileUrl`) or encodes itself (`dataURL`), but for a prefilled file with neither field it has no independent signal to validate `mimeType`/`ext`/`fileSize` against. Defaulting `trustProvidedFileMetadata: false` means: we don't silently present unverifiable metadata as trusted. Consumers that control their own pipeline and accept that trade-off can opt in with `true`.

src/context-providers/yup/helper.ts
```diff
@@ -36,30 +37,7 @@ export namespace YupHelper {
...
-		return Yup.object().meta({ yupId, whenDependencyMap }).shape(yupSchema, whenPairIds);
-	};
-
-	/**
-	 * Converts [dependentField, sourceField] pairs into a sourceField -> dependentFields[] map.
-	 * Used to determine which fields should revalidate when a source field changes.
-	 * @param whenPairIds array of [dependentFieldId, sourceFieldId] pairs
-	 * @returns map of sourceFieldId -> dependentFieldIds[]
-	 */
-	const buildWhenDependencyMap = (whenPairIds: [string, string][]): Record<string, string[]> => {
```
qroll, line 52: why was this removed?

shengxi-gt: MB, let me go through again and add that back

src/components/fields/image-upload/image-upload.tsx
```diff
@@ -155,7 +155,7 @@ export const ImageUploadInner = (props: IGenericFieldProps<IImageUploadSchema>)
 						);
 					}
 				),
-			validation
+			validation?.filter((rule) => !("matches" in rule))
```
qroll, line 158: why filter matches rule here?

qroll:
> > would argue against this change, the value are provided by the client/server in the first place
> 
> The concern isn't distrust of the originating server — it's that between the server delivering the prefilled payload and the browser submitting the form, the data lives in the page where a user (or MITM proxy) can modify it. The frontend can verify files it fetches (`fileUrl`) or encodes itself (`dataURL`), but for a prefilled file with neither field it has no independent signal to validate `mimeType`/`ext`/`fileSize` against. Defaulting `trustProvidedFileMetadata: false` means: we don't silently present unverifiable metadata as trusted. Consumers that control their own pipeline and accept that trade-off can opt in with `true`.

`trustProvidedFileMetadata`: if we're looking at prefilled data flow, `fileUrl` and `dataURL` also come from the same "server", what makes `uploadResponse` less trustworhty?

`allowedFileOrigins`: and then back to trusting `fileUrl`, shouldn't CSP `connect-src` help with restricting what origins are callable?

src/components/fields/image-upload/image-upload.tsx
qroll: ah ok figured out that this is to omit the rule from the yup-helper to avoid the warning since there's a separate treatment in the image upload manager

qroll:
> > jest just seems to hang indefinitely on this test case, this time check doesn't seem useful
> 
> You were right that the old version hung — the query was passed directly to `new RegExp()` without escaping. In this PR we added `escapeRegExp(query)` before building the `RegExp` (line 12 of `helper.ts`), so `(a+)+$` becomes the literal pattern `\(a\+\)\+\$`. No backtracking; the test completes in microseconds. We also moved the spec file to the `location-modal/location-search/` subdirectory as part of the reorganisation.

sorry what I meant was that without the fix, the test would hang instead of timing out. how can we make jest recover from this?

shengxi-gt:
> > > would argue against this change, the value are provided by the client/server in the first place
> > 
> > 
> > The concern isn't distrust of the originating server — it's that between the server delivering the prefilled payload and the browser submitting the form, the data lives in the page where a user (or MITM proxy) can modify it. The frontend can verify files it fetches (`fileUrl`) or encodes itself (`dataURL`), but for a prefilled file with neither field it has no independent signal to validate `mimeType`/`ext`/`fileSize` against. Defaulting `trustProvidedFileMetadata: false` means: we don't silently present unverifiable metadata as trusted. Consumers that control their own pipeline and accept that trade-off can opt in with `true`.
> 
> `trustProvidedFileMetadata`: if we're looking at prefilled data flow, `fileUrl` and `dataURL` also come from the same "server", what makes `uploadResponse` less trustworhty?
> 
> `allowedFileOrigins`: and then back to trusting `fileUrl`, shouldn't CSP `connect-src` help with restricting what origins are callable?

`trustProvidedFileMetadata`: Actually now that i think about it more, validating this should always be server-side anyways as we should always regard the browser as an untrusted environment. So this feature actually just increases friction without really achieving what it was intended for. Hence will remove it.

`allowedFileOrigins`: Point taken, will add a note to use CSP `connect-src` instead of building redundancy. Will remove to reduce upgrade friction.

shengxi-gt:
> sorry what I meant was that without the fix, the test would hang instead of timing out. how can we make jest recover from this?

we could try using vm.runInNewContext in this context for the test?

src/utils/style-helper.ts
```diff
@@ -0,0 +1,8 @@
+export namespace StyleHelper {
+	/** strips the two concrete network-side-channel risks from a schema-authored CSS string before it's
+	 * assigned to an element's style.cssText: @import (loads a remote stylesheet) and url() (references
+	 * a remote resource). Not a full CSS parser/allowlist — legitimate uses of this styling hook
+	 * (padding/margin tweaks) need neither construct. */
+	export const sanitizeStyleString = (value: string): string =>
+		value.replace(/@import[^;]*;?/gi, "").replace(/url\s*\([^)]*\)/gi, "");
```
weili-govtech, line 7: this will silently strip away valid use of background images, do update the corresponding prop descriptions in image-upload / location-field stories and add a changelog note of this breaking change

src/context-providers/yup/helper.ts
shengxi-gt: Resolved, https://github.com/LifeSG/web-frontend-engine/pull/739/commits/4ad599b69fd832a4fd4383cb4645b8ad57200d3c

src/utils/style-helper.ts
shengxi-gt: Resolved, https://github.com/LifeSG/web-frontend-engine/pull/739/commits/e4329bbaa75f1366ad4435f9ac4d07ba1001dd7a

weili-govtech (dismissed):

src/stories/3-fields/location-field/location-field.stories.tsx
```diff
@@ -186,9 +186,20 @@ const meta: Meta = {
...
+				“Prevents confirming and submitting locations that are outside Singapore. On confirming any selected location — a searched address (e.g. `CAUSEWAY (JOHOR)`), a map selection or an unresolvable `Pin location: <lat>, <lng>` value — its coordinates are checked against the coastal outlines of SLA's National Map Polygon dataset: if it falls on a neighbouring (JOHOR (MALAYSIA)) landmass or in waters outside Singapore, the “This location is outside Singapore.” prompt is shown and the location modal stays open. Areas within Singapore that simply have no addresses nearby (e.g. sea just off the coast, reservoirs) remain confirmable. Prefilled values that resolve to locations outside Singapore are cleared and such values fail validation on submission.”,
```
qroll: the quote characters in this file are invalid, will break the storybook build

src/__tests__/components/fields/location-field/location-modal/location-search/helper.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+	it("should treat regex metacharacters in the query as literal characters", () => {
+		const [result] = boldResultsWithQuery([buildResult("Blk 5 (Example)")], "(Example)");
+
+		expect(result.displayAddressText).toBe('Blk 5 <span class="keyword">(Example)</span>');
+	});
+
+	it("should complete within a reasonable time for any query string", () => {
+		const input = [buildResult("a".repeat(25) + "!")];
+		expect(() =>
+			vm.runInNewContext("fn(input, query)", { fn: boldResultsWithQuery, input, query: "(a+)+$" }, { timeout: 1000 })
```
qroll: weirdly the non-escaped regex still resolves in 1s on my machine. setting the string length to 30 works for me

```suggestion
		const input = [buildResult("a".repeat(30) + "!")];
		expect(() =>
			vm.runInNewContext("fn(input, query)", { fn: boldResultsWithQuery, input, query: "(a+)+$" }, { timeout: 1000 })
```

qroll (changes requested):
the unit tests are stuck when I run them locally. do you encounter this too?

qroll (approved):

PR #740, opened 2026-09-18
Hardening pass: input validation, origin checks, and default-safe field behaviour — legacy/v2

shengxi-gt:
# Hardening pass — legacy/v2

Same changes as the `main` branch PR, adapted for v2's codebase (`Text` instead of `Typography`, styled-components CSS interpolation instead of direct `cssText`).

## ⚠️ Breaking changes

Four changes alter default runtime behaviour unconditionally, no opt-out, for every existing schema as soon as this ships:

1. **`Iframe` `postMessage` origin check is now hard default-on.** Messages from a mismatched origin are silently dropped.
2. **1000-character cap on schema-authored regex input.** Inputs longer than 1000 chars are treated as non-matching regardless of the actual pattern result. Affects `matches`, `notMatches`, and `filenameMatches` rules.
3. **MaskedField `maxLength` now defaults to 500 whenever `maskRegex` is set** and no `max`/`length` rule already narrows it.
4. **Pattern strings not wrapped in `/pattern/flags` slashes are no longer silently skipped.** They previously landed in a catch block and validation was skipped entirely (value always passed). The shared parsing helper now falls back to treating the whole string as a raw pattern and applies it — any schema relying on the old silent no-op will now get real validation applied.

Narrower breaking changes (only affect a consumer that relied on the specific permissive behaviour being fixed): stricter `sanitize-html` configuration may now strip attributes that previously passed through; the ButtonField URL scheme check may block non-standard schemes; custom modal style strings now have `@import`/`url()` stripped.

## Changes covered

- `sanitize-html` configuration tightened in Text, FilterCheckbox, Popover
- Iframe `postMessage` now validates origin by default
- Location-field search query now escaped before use in `RegExp`
- OTP `state` field documented as client-asserted (no code change)
- Schema regex inputs now capped at 1000 characters
- MaskedField `maxLength` defaults to 500 when `maskRegex` is set
- Custom modal style strings now strip `@import` and `url()` references
- ImageUpload `matches` rule no longer blocks valid submissions (pre-existing bug fix)

## Test plan

- Full Jest suite + `eslint` + `tsc --noEmit` pass on this branch.


shengxi-gt:
You were right that the old version hung — the query was passed directly to `new RegExp()` without escaping. In this PR we added `escapeRegExp(query)` before building the `RegExp` (line 12 of `helper.ts`), so `(a+)+$` becomes the literal pattern `\(a\+\)\+\$`. No backtracking; the test completes in microseconds. We also moved the spec file to the `location-modal/location-search/` subdirectory as part of the reorganisation.

shengxi-gt:
The concern isn't distrust of the originating server — it's that between the server delivering the prefilled payload and the browser submitting the form, the data lives in the page where a user (or MITM proxy) can modify it. The frontend can verify files it fetches (`fileUrl`) or encodes itself (`dataURL`), but for a prefilled file with neither field it has no independent signal to validate `mimeType`/`ext`/`fileSize` against. Defaulting `trustProvidedFileMetadata: false` means: we don't silently present unverifiable metadata as trusted. Consumers that control their own pipeline and accept that trade-off can opt in with `true`.

qroll (dismissed):

qroll (approved):

